PR #27, opened 2025-04-02
feat: Td 202 zeichenbegrenzungen in allen input felder

JoshuaZiegler:
This PR assigns all configuration text-fields a maximum input size, exceeding the limit will prevent all further inputs


nachthammer (approved):
i thought that you cannot write more input characters and that it is not only done via zod validation, i am not sure about the ticket description tho, there has to be some kind of feedback, if you exceed the length

PR #88, opened 2025-06-04
feat: TD- 23 add images to chat

JoshuaZiegler:


src/app/api/chat/route.ts
```diff
@@ -130,6 +132,13 @@ export async function POST(request: NextRequest) {
 
   const orderedChunks = await getRelevantFileContent({ messages, user, relatedFileEntities });
 
+  // Extract images from attached files and convert to base64
```
recursive-hub, line 135: This comment seems outdated since the images are not converted to base64 .

JoshuaZiegler: yes, it's outdated

recursive-hub (approved):

PR #100, opened 2025-06-12
fix: bug for manadatory form fields

JoshuaZiegler:
add: test for form validation

### DIFF
--- a/e2e/tests/create-custom-gpt.test.ts
+++ b/e2e/tests/create-custom-gpt.test.ts
@@ -168,3 +168,36 @@ test('data is autosaved on blur', async ({ page }) => {
   await page.getByPlaceholder('Erstelle einen').press('Tab');
   await expect(page.getByPlaceholder('Erstelle einen')).toHaveValue('New Prompt Suggestion');
 });
+
+test('submit button is disabled when required fields are empty', async ({ page }) => {
+  await login(page, 'teacher');
+
+  await page.goto('/custom');
+  await page.waitForURL('/custom');
+
+  const createButton = page.getByRole('button', { name: 'Assistent erstellen' });
+  await expect(createButton).toBeVisible();
+  await createButton.click();
+
+  await page.waitForURL('/custom/editor/**');
+
+  const submitButton = page.getByRole('button', { name: 'Assistent erstellen' });
+  await expect(submitButton).toBeVisible();
+
+  // Check that submit button is disabled when all fields are empty
+  await expect(submitButton).toBeDisabled();
+
+  // Fill in only the name - button should still be disabled due to missing description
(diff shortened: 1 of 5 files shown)


e2e/tests/create-custom-gpt.test.ts
```diff
@@ -168,3 +168,36 @@ test('data is autosaved on blur', async ({ page }) => {
...
+  await expect(submitButton).toBeVisible();
+
+  // Check that submit button is disabled when all fields are empty
+  await expect(submitButton).toBeDisabled();
+
+  // Fill in only the name - button should still be disabled due to missing description
+  await page.getByRole('textbox', { name: 'Wie soll diese' }).fill('Test GPT Name');
+  await expect(submitButton).toBeDisabled();
+
+  // Fill in the required description field - now button should be enabled
```
xit-paul-reinhardt, line 194: This comment does not match the assertion in line 199. Should the button still be disabled or now enabled?

src/app/(authed)/(dialog)/characters/editor/[characterId]/character-form.tsx
```diff
@@ -86,12 +89,12 @@ export default function CharacterForm({
     resolver: zodResolver(characterFormValuesSchema),
     defaultValues: {
       ...character,
-      description: character.description ?? '',
-      learningContext: character.learningContext ?? '',
-      competence: character.competence ?? '',
-      schoolType: character.schoolType ?? '',
-      gradeLevel: character.gradeLevel ?? '',
-      subject: character.subject ?? '',
+      description: character.description ?? null,
```
Comment (xit-paul-reinhardt) [inline src/app/(authed)/(dialog)/characters/editor/[characterId]/character-form.tsx:92]: Why are these fallbacks `null` and the others below `undefined`? Is this an accident or is there an intended difference?

src/app/(authed)/(dialog)/characters/editor/[characterId]/character-form.tsx
```diff
@@ -47,11 +47,14 @@ type CharacterFormProps = CharacterModel & {
   readOnly: boolean;
 };
 
+/**
+ * If the field is required, set the min length to at least 1.
```
Comment (xit-paul-reinhardt) [inline src/app/(authed)/(dialog)/characters/editor/[characterId]/character-form.tsx:51]: We now have this comment in some, but not all schemas containing required fields (like `customGptFormValuesSchema`) I would propose to either document it in a more central place like a short description on how we use zod or to add the comment to all schemas containing required fields.

xit-paul-reinhardt (changes requested):

src/app/(authed)/(dialog)/characters/editor/[characterId]/character-form.tsx
Comment (JoshuaZiegler) [reply src/app/(authed)/(dialog)/characters/editor/[characterId]/character-form.tsx]: It can all be undefined everwhere, it appears to conflict with the inferred type from drizzle, if you want to set it to null

xit-paul-reinhardt (dismissed):

PR #126, opened 2025-07-29
DE-BW Disclaimer

recursive-hub:
- added image capability to the generic disclaimer generator
- the checkbox label is now also Markdown, since we need to display a link for DE-BW
- added DE-BW disclaimer config
- removed ~60000 lines of unused user-mappings and added 2 useful users
- added image as static asset


apps/dialog/e2e/fixtures/user-mappings.json
rlaerm, line 1: It seems that you have deleted most of the users? Are you sure that we do not need them any longer for the load tests?

recursive-hub: Yes. This is a file that we edit manually from time to time, so I want to have it as clean as possible. If we need auto generated users for a future load test, then we would add a new file with auto generated users.

apps/dialog/public/disclaimer/DE-BW.png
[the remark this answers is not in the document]
recursive-hub: done

PR #157, opened 2025-09-12
TD-240 add visible textbox limits

AsamMax:


apps/dialog/src/components/common/text-input.tsx
```diff
@@ -44,12 +44,23 @@ export function TextInput<T extends HTMLTextAreaElement | HTMLInputElement>({
 
   const effectiveMaxLength: number | undefined = props.maxLength ?? defaultMaxLength;
   const [invalid, setInvalid] = React.useState(false);
+  const [counterVisible, setCounterVisible] = React.useState(false);
+  const [liveValue, setLiveValue] = React.useState(getValue());
```
rlaerm, line 48: What is a liveValue? What about textLength, characterCount, currentLength or something like that?

apps/dialog/src/components/common/text-input.tsx
```diff
@@ -78,45 +89,79 @@ export function TextInput<T extends HTMLTextAreaElement | HTMLInputElement>({
...
         {inputType === 'textarea' ? (
-          <textarea
-            className={cn(
-              'w-full',
-              inputFieldClassName,
-              'focus:border-primary placeholder:text-gray-300',
-              customInputClassName,
-              'resize-none',
-              borderErrorClass,
+          <React.Fragment>
```
rlaerm, line 104: You can also use <> if you like

rlaerm (approved):

PR #717, opened 2026-03-31
Refactor header component

rlaerm:


rlaerm:
Rejected because it does not work as expected.

PR #770, opened 2026-04-13
TD-1110 use storage state for persistent login in e2e tests

TobiasWallura-xitaso:


apps/dialog/e2e/global-setup.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default async function globalSetup(config: FullConfig) {
+  const baseUrl = config.projects[0]?.use.baseURL;
+  if (!baseUrl) {
+    throw new Error('Playwright globalSetup requires `use.baseURL` to be configured. ');
+  }
+
+  await using browser = await chromium.launch();
+
+  for (const [user, file] of Object.entries(AUTH_FILES)) {
```
rlaerm, line 29: I have not tested it yet but I assume that I am not able to run the e2e tests any longer in my local env, right? I have to create those files otherwise the tests are not able to login a teacher. If there is no file we should throw an error to remind the user to create those files.

apps/dialog/e2e/global-setup.ts
```diff
@@ -0,0 +1,26 @@
...
+    await fs.mkdir(path.dirname(file), { recursive: true });
+
+    await using context = await browser.newContext({ baseURL: baseUrl });
+    const page = await context.newPage();
+
+    await login(page, user);
+    await context.storageState({ path: file });
```
rlaerm, line 36: I do not really understand this. Maybe you can introduce this feature.

rlaerm (approved):

apps/dialog/e2e/global-setup.ts
TobiasWallura-xitaso: described in the jsdoc: global setup is run automatically by playwright, so it works without manual intervention.

PR #774, opened 2026-04-14
Feat/td 1104 design config

rlaerm:


packages/shared/migrations/0059_small_mephisto.sql
TobiasWallura-xitaso, line 1: Does this have a big effect on the old design? if yes, we might want to do a deployment before.

TobiasWallura-xitaso (approved):

packages/shared/migrations/0059_small_mephisto.sql
rlaerm: Hopefully it is safe to execute that migration immediately. I checked the old ui with settings from Saarland and it looks fine to me.

PR #777, opened 2026-04-14
Minor Styling and Accessibility adjustments for Entity Overview

AsamMax:


apps/dialog/src/components/entity-overview/entity-overview.tsx
TobiasWallura-xitaso, line 109: When do we change this to our `Button` component?

TobiasWallura-xitaso (approved):

apps/dialog/src/components/entity-overview/entity-overview.tsx
AsamMax: When / If we want to have a special type for this exact case. It has an odd size, pretty special placement and styling that is completely seperate from all existing cases

PR #804, opened 2026-04-16
Td 1119 remove pictureid handling in frontend

brueckb:


apps/admin/src/app/telli-dialog/templates/templateTypeName.ts
```diff
@@ -4,7 +4,7 @@ export function getTemplateTypeName(type: TemplateModel['type']): string {
   switch (type) {
     case 'character':
       return 'Dialogpartner';
-    case 'custom-gpt':
```
rlaerm, line 7: There is another file that must be changed. It is not directly part of the story but related and broken atm. If you look at "CreateTemplateModal.tsx" there is a check if the pasted url starts with "custom|characters|learnings-scenarios" which is not correct any longer. Replate custom with assistants please. Thx.

rlaerm (approved):

packages/shared/src/learning-scenarios/learning-scenario-admin-service.ts
```diff
@@ -1,24 +1,23 @@
...
-import { buildLearningScenarioPictureKey } from '@shared/learning-scenarios/learning-scenario-service';
-import { copyFileInS3 } from '@shared/s3';
-import { duplicateFileWithEmbeddings } from '@shared/files/fileService';
 import { and, eq, lt } from 'drizzle-orm';
 import { addDays } from '@shared/utils/date';
-import path from 'node:path';
+import {
+  copyEntityPictureIfExists,
+  copyRelatedTemplateFiles,
+} from '@shared/templates/template-service';
```
AsamMax, line 19: This might cause a circular import, right?

template-service.ts → learning-scenario-service.ts
learning-scenario-service.ts → learning-scenario-admin-service.ts
learning-scenario-admin-service.ts → template-service.ts ← new

packages/shared/src/templates/template-service.test.ts
AsamMax: I love nitpicking :). 
```suggestion
    it('should copy and link files for assistant templates', async () => {
```

AsamMax (approved):

packages/shared/src/templates/template-service.test.ts
AsamMax: No idea what went wrong with the suggestion. Obviously custom-gpt → assistant

apps/admin/src/app/telli-dialog/templates/templateTypeName.ts
brueckb: great catch! will do :)

packages/shared/src/learning-scenarios/learning-scenario-admin-service.ts
brueckb: added a small picture-key utils file to break the circular dependency

PR #813, opened 2026-04-17
Feat/td 1114 resolve test findings

rlaerm:


apps/dialog/src/app/(authed)/(dialog)/learning-scenarios/[learningScenarioId]/learning-scenario-view.tsx
```diff
@@ -122,7 +122,7 @@ export function LearningScenarioView({
           </CardContent>
         </Card>
 
-        {learningScenario.accessLevel === 'global' && (
+        {learningScenario.accessLevel !== 'global' && (
```
Comment (AsamMax) [inline apps/dialog/src/app/(authed)/(dialog)/learning-scenarios/[learningScenarioId]/learning-scenario-view.tsx]: Only learning scenarios check !== 'global'. This seems wrong. Isn't it now displaying FWU as Author for school shared Learning Scenarios?

apps/dialog/src/components/custom-chat/custom-chat-layout-container.tsx
```diff
@@ -1,3 +1,5 @@
 export function CustomChatLayoutContainer({ children }: { children: React.ReactNode }) {
-  return <div className="relative pb-8 flex flex-col gap-6">{children}</div>;
+  return (
+    <div className="data-custom-chat-layout relative flex flex-col gap-6 pb-8">{children}</div>
```
AsamMax, line 3: What does the 'data-custom-chat-layout' css class do? can't find anything in the tailwind documentation.

apps/dialog/src/components/layout/default-page-layout.tsx
```diff
@@ -5,5 +5,5 @@ type DefaultPageLayoutProps = {
 };
 
 export function DefaultPageLayout({ children }: DefaultPageLayoutProps) {
-  return <div className="max-w-5xl mx-auto px-6 h-full">{children}</div>;
+  return <div className="data-page-layout h-full max-w-5xl mx-auto px-6 pb-8">{children}</div>;
```
AsamMax, line 8: Same thing here, what does 'data-page-layout' do?

AsamMax, line 8: Does adding a pb-8 here collide / stack with already existing paddings on individual pages? For example the overview page has a pb-6 on the last element.

rlaerm: Sadly it behaves different for pages that define a scrollable area inside the container. But because of h-full it does not stack.

rlaerm: That is just a nice method to make it easier to identify those divs in the DOM. It does not modify anything.

apps/dialog/src/app/(authed)/(dialog)/learning-scenarios/[learningScenarioId]/learning-scenario-view.tsx
Comment (rlaerm) [reply apps/dialog/src/app/(authed)/(dialog)/learning-scenarios/[learningScenarioId]/learning-scenario-view.tsx]: Oh no, that is a testing mistake. My bad.

apps/dialog/src/components/layout/default-page-layout.tsx
AsamMax: image
Actually it seems it does stack. Both pb create space at the bottom, totalling something like a "pb-14". the default page pb-8 even "cuts off" the bottom here visually.

AsamMax: But i can't find a test where it is used to identify it either. For now this seems a bit superfluous. I'd rather see such things added when needed.

AsamMax (approved):

apps/dialog/src/components/layout/default-page-layout.tsx
rlaerm: Okay, it does stack in the overview but does not in all the others. I recognized that your solution looks better because the scrollbar is at the right border of the window. That interfers with the idea of the DefaultPageLayout. Maybe I can adapt it but there is a scrollRef in it. Let me check that.

rlaerm: Okay, now we use the main scrollable area in all pages consistently. And the scroll to top thing is still working.

PR #826, opened 2026-04-22
Info Banners

AsamMax:


packages/shared/src/db/schema.ts
```diff
@@ -287,6 +288,96 @@ export type FederalStateSelectModel = z.infer<typeof federalStateSelectSchema>;
...
+export type InfoBannerType = z.infer<typeof infoBannerTypeSchema>;
+
+export const infoBannerTable = pgTable(
+  'info_banner',
+  {
+    id: uuid('id').defaultRandom().primaryKey(),
+    type: infoBannerTypeEnum('type').notNull(),
+    message: text('message').notNull(),
+    ctaLabel: text('cta_label'),
+    ctaUrl: text('cta_url'),
```
TobiasWallura-xitaso: What's `cta`?

AsamMax: Cta means call to action.

But good point, i'll change this to just button_*

packages/shared/src/db/functions/vidis.ts
```diff
@@ -60,8 +61,12 @@ export async function dbGetOrCreateVidisUser(userInfo: VidisUserInfo) {
           firstName: '',
           lastName: '',
           email: `${userInfo.sub}@vidis.schule`,
+          loginCount: 1,
+        })
+        .onConflictDoUpdate({
+          target: [userTable.id],
+          set: { loginCount: sql`${userTable.loginCount} + 1` },
```
TobiasWallura-xitaso: Do I understand this correct: on each login with vidis, the `loginCount` is incremented. This is not related to any displayed info banners. How can this loginCount then be used for comparison with the max login count of the info banner?
Don't we need to store the count how many times a specific info banner was already shown to the user?

apps/dialog/src/components/info-banners/active-info-banners.tsx
```diff
@@ -0,0 +1,116 @@
...
+          {currentInfoBanner.ctaLabel && currentInfoBanner.ctaUrl ? (
+            <Button asChild size="sm" className="shrink-0">
+              <Link href={currentInfoBanner.ctaUrl} target="_blank" rel="noopener noreferrer">
+                {currentInfoBanner.ctaLabel}
+              </Link>
+            </Button>
+          ) : null}
+        </div>
+        <div className="absolute right-0 top-1/2 flex -translate-y-1/2 items-center justify-center gap-1">
+          <button
```
TobiasWallura-xitaso: Prefer to use our Button component.

packages/shared/src/info-banners/info-banner-service.ts
```diff
@@ -0,0 +1,213 @@
...
+    )
+    .where(
+      and(
+        eq(infoBannerFederalStateMappingTable.federalStateId, federalStateId),
+        eq(infoBannerTable.isDeleted, false),
+        lte(infoBannerTable.startsAt, now),
+        gt(infoBannerTable.endsAt, now),
+        or(
+          isNull(infoBannerTable.maxLoginCount),
+          sql`${infoBannerTable.maxLoginCount} >= ${loginCount}`,
```
brueckb: i second this comment:  https://github.com/FWU-DE/telli-dialog/pull/826#discussion_r3125118896. if i understand it correctly, a user that has logged in 5 times since the implementation won't see a newly created info-banner with maxLoginCount of 4 :)

brueckb (approved):

PR #848, opened 2026-04-29
Rebranding

elenamahartana:


recursive-hub (approved):

brueckb (approved):

apps/dialog/src/app/layout.tsx
```diff
@@ -14,7 +14,7 @@ import { buildPublicConfig } from '@shared/sentry/public-config';
 import { cn } from '@/utils/tailwind';
 import { getReadOnlySignedUrl } from '@shared/s3';
 import { SEVEN_DAYS } from '@shared/s3/const';
-import telliIcon from '@/assets/telli.svg';
+import logoIcon from '@/assets/logo-only.svg?url';
```
TobiasWallura-xitaso, line 17: What does `?url` do?
*Edit*: it loads it as React component, as configured in next config. What's the benefit of that? also the lib `@svgr/webpack` was last published 3 years ago.

TobiasWallura-xitaso (approved):

apps/dialog/src/app/layout.tsx
rlaerm: Normally svgs are loaded as react component via svgr. With the ?url parameter I am allowed to load the svg as file. That is necessary for favicon or the manifest file. In all other cases we can just import svgs without any parameters.

PR #849, opened 2026-04-29
refactor authentication-service and error handling with tests

rlaerm:


apps/dialog/messages/de.json
```diff
@@ -481,7 +481,9 @@
   },
   "authentication": {
     "login-error": {
-      "description": "Login nicht möglich. Folgende Informationen müssen mit dem Profil übertragen werden:"
+      "description": "Login nicht möglich. Folgende Informationen müssen mit dem Profil übertragen werden:",
+      "federal-state-not-found": "Das übermittelte Bundesland wurde nicht gefunden.",
+      "federal-state-changed": "Das übermittelte Bundesland stimmt nicht mit den vorhandenen Nutzerdaten überein."
```
TobiasWallura-xitaso: Didn't we want to support changing the federal state?
```suggestion
```

rlaerm: Yes, you are right.

packages/shared/src/db/functions/user.ts
```diff
@@ -42,3 +42,65 @@ export async function dbUpdateUserTermsVersion({
...
+    schoolIds: string[];
+    federalStateId: string;
+    userRole: UserRole;
+  },
+) {
+  const [insertedUser] = await db.insert(userTable).values(user).returning();
+  return insertedUser;
+}
+
+export async function dbUpdateUserById(
```
TobiasWallura-xitaso, line 58: Do we need this fn, when we have `dbCreateOrUpdateUser`?

TobiasWallura-xitaso (approved):

PR #853, opened 2026-04-30
TD-1180: fix three dot menu

stschaller-xitaso:


apps/dialog/src/components/navigation/profile-menu.tsx
```diff
@@ -9,19 +9,18 @@ import Link from 'next/link';
 import { IMPRESSUM_URL, PRIVACY_POLICY_URL, TERMS_OF_USE_URL } from './const';
 import { useTranslations } from 'next-intl';
 import DotsHorizontalIcon from '@/components/icons/dots-horizontal';
```
TobiasWallura-xitaso: If the file is now unused, remove that file as well.

TobiasWallura-xitaso (approved):

PR #874, opened 2026-05-06
Introduce Billing for Web Search

AsamMax:


packages/shared/src/tool-call-costs/tool-call-cost-service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return toolCallCostSchema.parse(toolCallCost);
+}
+
+export async function updateToolCallCost(input: UpdateToolCallCostInput): Promise<ToolCallCost> {
+  const values = updateToolCallCostSchema.parse(input);
+  const updatedToolCallCost = await dbUpdateToolCallCost(values);
+
+  return toolCallCostSchema.parse(updatedToolCallCost);
+}
```
TobiasWallura-xitaso, line 25: Why do we parse the data with zod after reading from database?

packages/shared/src/tool-call-costs/tool-call-cost.ts
TobiasWallura-xitaso, line 1: This file looks like validation and labels for the admin ui. Is the shared package the appropriate place for it?

rlaerm:
test comment

packages/shared/src/db/schema.ts
```diff
@@ -140,7 +143,7 @@ export const conversationMessageTable = pgTable(
     createdAt: timestamp('created_at', { mode: 'date', withTimezone: true }).defaultNow().notNull(),
     deletedAt: timestamp('deleted_at', { mode: 'date', withTimezone: true }),
     parameters: json('parameters').$type<ConversationMessageParameters>(),
-    webSearchResults: json('web_search_results').$type<TextSearchResult[]>(),
+    webSearchResults: json('web_search_results').$type<WebSearchResult[]>(),
   },
```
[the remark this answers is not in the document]
TobiasWallura-xitaso: Usually we depend on the type from the linkup sdk. Please double check with @recursive-hub

packages/shared/src/db/types.ts
```diff
@@ -50,6 +50,14 @@ export type WebsearchSource = {
   error?: boolean;
 };
 
+export type WebSearchResult = {
+  type: 'text';
+  name: string;
+  url: string;
+  content: string;
+  favicon: string;
+};
```
TobiasWallura-xitaso: also see other comment, this type can be removed.

packages/shared/src/db/functions/tool-call.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function dbGetToolCallCosts() {
+  return db.select().from(toolCallCostTable).orderBy(asc(toolCallCostTable.toolCallName));
+}
+
+export async function dbGetToolCallCostByName(toolCallName: ToolCallName) {
+  const [toolCallCost] = await db
+    .select()
+    .from(toolCallCostTable)
+    .where(eq(toolCallCostTable.toolCallName, toolCallName));
```
TobiasWallura-xitaso, line 13: If these queries are used in prod code, they could be cached for performance.

packages/shared/src/db/schema.ts
```diff
@@ -735,14 +738,51 @@ export type LearningScenarioOptionalShareDataModel = z.infer<
...
 /**
  * Schema for table shared_learning_scenario_usage_tracking
  */
 export const sharedLearningScenarioUsageTracking = pgTable(
   'shared_learning_scenario_usage_tracking',
   {
     id: uuid('id').defaultRandom().primaryKey(),
-    modelId: uuid('model_id').notNull(),
+    modelId: uuid('model_id'),
+    toolCallName: toolCallNameEnum('tool_call_name'),
```
TobiasWallura-xitaso, line 791: How do we save the tool call costs for a row in this usage tracking table?

TobiasWallura-xitaso: same for all of the other tables.

apps/admin/src/app/telli-dialog/tool-call-costs/ToolCallCostListView.tsx
```diff
@@ -0,0 +1,199 @@
...
+    toolCallCosts.map((toolCallCost) => [
+      toolCallCost.toolCallName,
+      String(toolCallCost.costsInCent),
+    ]),
+  );
+}
+
+export default function ToolCallCostListView() {
+  const [toolCallCosts, setToolCallCosts] = useState<ToolCallCost[]>([]);
+  const [draftCosts, setDraftCosts] = useState<Record<string, string>>({});
```
TobiasWallura-xitaso: Why do we need a draft state? Also I'd prefer to use `react-hook-form`.
From my point of view, the form does not need to be generic to support any number of tools, as we only have a single tool right now.

packages/shared/src/db/schema.ts
TobiasWallura-xitaso: So the logic is: when `toolCallName` is set, the costs are only for the tool call. same for the `modelId`, when that is set, the costs are only for the model call. Is that correct?

AsamMax: Yes, it was simpler this way after all. This way we can easily extract only llm costs (those with a model) and the sum stays consistent

TobiasWallura-xitaso (approved):

PR #877, opened 2026-05-06
Web search renaming

recursive-hub:
- renamed WebSearchSource to WebSource, since it is not used in a web search context but rather as an URL container, for the Citation component and as result of a web scrape. I will see if I can further clean up this object, since it is used for different purposes
- renamed webSearch to webScrape, in case it is not a real search


brueckb (approved):

apps/dialog/src/app/(authed)/(dialog)/characters/[characterId]/page.tsx
```diff
@@ -23,7 +23,7 @@ export default async function Page(props: PageProps<'/characters/[characterId]'>
         ({
           link: url,
           error: false,
-        }) as WebsearchSource,
+        }) as WebSource,
```
Comment (rlaerm) [inline apps/dialog/src/app/(authed)/(dialog)/characters/[characterId]/page.tsx:26]: Why do you claim it is a WebSource? That should not be necessary, could be dangerous and is not helpfull in my opinion.

apps/dialog/src/app/api/chat/websearch.ts
rlaerm, line 1: I ask myself why this file is located under dialog/.../api/chat
I would create a new folder in the shared package where our business logic resides. There should be a websearch-service as entry point. Maybe there is a component 'websearch-decider' or 'websearch-agent' or something like this. For me, it is hard to understand the structure. If you think about our documentation, where to put this in the building block view and which dependencies do you have?

apps/dialog/src/app/api/utils/extract-urls.ts
rlaerm, line 1: Why is this file located under dialog/.../api/utils? Web-scraping should be a unit/component/service like web-search.

apps/dialog/src/configuration-text-inputs/const.ts
```diff
@@ -1,20 +1,32 @@
...
 export const TEXT_INPUT_FIELDS_LENGTH_LIMIT = 500;
 export const TEXT_INPUT_FIELDS_LENGTH_LIMIT_FOR_DETAILED_SETTINGS = 10000;
 export const CHAT_MESSAGE_LENGTH_LIMIT = 6000;
 export const SMALL_TEXT_INPUT_FIELDS_LIMIT = 50;
 export const TOTAL_CHAT_LENGTH_LIMIT = 60_000; // Maximum total characters in chat history (~40 DIN A4 pages)
-export const TOTAL_WEBSEARCH_CONTENT_LENGTH_LIMIT = 30_000; // Maximum total characters from websearch content (~20 DIN A4 pages)
-export const SINGLE_WEBSEARCH_CONTENT_LENGTH_LIMIT = 6000; // Maximum total characters from a single websearch content (~4 DIN A4 pages)
+
+// Web scraper
+export const WEB_SCRAPE_RESULT_LENGTH_LIMIT = 6000; // Maximum total characters from a single web scrape result (~4 DIN A4 pages)
```
rlaerm, line 9: The file is about consts in **configuration-text-input**. I would not expect config values for web-scrape here.

rlaerm: Oh, I see there are a lot of consts that do not belong to text-inputs. Quick fix would be to put the file into another folder. But maybe we should discuss if we like a solution where all config values are in the same place or multiple config files where they belong to like we do with the env variables.

apps/dialog/src/app/(authed)/(dialog)/characters/[characterId]/page.tsx
Comment (recursive-hub) [reply apps/dialog/src/app/(authed)/(dialog)/characters/[characterId]/page.tsx:26]: You are right, this is a legacy artefact. I will see if I can further slipt up the object WebSource, since it is sometimes just used as an url wrapper.

apps/dialog/src/app/api/chat/websearch.ts
recursive-hub: Good idea, I will move it in the next pr

apps/dialog/src/app/api/utils/extract-urls.ts
recursive-hub: Will move it

apps/dialog/src/configuration-text-inputs/const.ts
recursive-hub: Yes lets discuss this

PR #879, opened 2026-05-06
TD-1199: fix hover effects and placeholder colors

stschaller-xitaso:


brueckb (approved):

apps/dialog/src/components/chat/sources/web-search-sources.tsx
```diff
@@ -139,7 +139,7 @@ export function WebSearchSourcesButton({
       aria-controls={panelId}
       variant={'ghost'}
       size={'icon'}
-      className="text-primary hover:text-primary hover:bg-secondary/30"
+      className="hover:bg-secondary/30"
```
rlaerm: Why do we have /30 here instead of /15? Could be worth a comment if we differ from our default

rlaerm (approved):

apps/dialog/src/components/chat/sources/web-search-sources.tsx
stschaller-xitaso: I asked Max and I'll remove it, the standard works fine here

PR #882, opened 2026-05-06
Setup agent skills in project

TobiasWallura-xitaso:
This pull request installs the Next.js best practices and shadcn skills in the project and adds a default AGENTS.md file.


brueckb (approved):

rlaerm:
To be honest, I have not a good feeling about this. Some of the best practices do not align with our current state. We do not use lucide ui library. First, we should align on the best practices we want to use. Second, does it really work? Maybe there are a lot of things described but in practice it does not work or is obsolete. In our last training we had the example, that there was a skill about accessibility with all the A, AA and AAA criterias in it. The other possibility was to put a semantic anchor like 'Web Content Accessibility Guidelines (WCAG) 2.2' in copilot instruction file. The generated output was nearly the same. Third, we should speak about the process. Which tool did you use to generate that. How to update and maintain that?
In general I like the idea and we can try that out. But maybe step by step and a small discussion first.

janrenz:
I would like to have a discussion on this at some point, we started also a fwu wide skill repo. so framework specific skills could be centralized or somehow synced

TobiasWallura-xitaso:
> To be honest, I have not a good feeling about this. Some of the best practices do not align with our current state. We do not use lucide ui library. First, we should align on the best practices we want to use. Second, does it really work? Maybe there are a lot of things described but in practice it does not work or is obsolete. In our last training we had the example, that there was a skill about accessibility with all the A, AA and AAA criterias in it. The other possibility was to put a semantic anchor like 'Web Content Accessibility Guidelines (WCAG) 2.2' in copilot instruction file. The generated output was nearly the same. Third, we should speak about the process. Which tool did you use to generate that. How to update and maintain that? In general I like the idea and we can try that out. But maybe step by step and a small discussion first.

My main intention was to get better results from copilot for our shadcn components. e.g. Copilot would often use normal `<button>` instead of our `Button` shadcn component.

The two skills were installed using the skills package (see https://www.npmjs.com/package/skills).

Totally agree to start a discussion, if the changes from this PR are the way to go :+1:

TobiasWallura-xitaso:
Replaced by #925

PR #893, opened 2026-05-08
Gently Reduce the frequency of suggestions

AsamMax:


apps/dialog/src/app/api/utils/system-prompt.ts
```diff
@@ -31,11 +31,11 @@ export const FORMAT_GUIDELINES = `
 
 export const SUGGESTION_GUIDELINES = `
 ## Vorschläge und Rückfragen
-Beende die Antwort, falls sinnvoll, mit einer passenden Rückfrage oder hilfreichen Vorschlägen, um den User zu inspirieren. 
+Solltest du passende Rückfragen oder hilfreiche Vorschläge haben, um den User zu inspirieren, beende die Antwort damit.
```
brueckb: feels like a veeery gentle adjustment to me. I am very curious about the impact, let's try it!

brueckb (approved):

TobiasWallura-xitaso (approved):

apps/dialog/src/app/api/utils/system-prompt.ts
elenamahartana: this lines feels very unnecessary. Especially as it often did#t work to answer with one letter.

apps/dialog/src/app/api/utils/system-prompt.ts
```diff
@@ -31,11 +31,11 @@ export const FORMAT_GUIDELINES = `
...
 Bei einfachen Fragen erstelle nur einen Vorschlag. Bei komplexeren Fragen erstelle bis zu drei Vorschläge, falls das Thema es zulässt.
```
elenamahartana, line 35: how about adding: Falls du im Chatverlauf bereits Vorschläge erteilt hast und diese nicht genutzt wurden, schlage nichts mehr vor.

apps/dialog/src/app/api/utils/system-prompt.ts
elenamahartana: maybe it would be nice if the model ist encouraged to answer shortly and as well as possible and only ask for clarification questions if it is really necessary.

elenamahartana: instead of "passende Rückfragen" I would prefer "notwendige Rückfragen".

PR #899, opened 2026-05-08
TD-1227: add default layout selector for correct padding in all pages

stschaller-xitaso:


apps/dialog/src/app/(authed)/(dialog)/characters/page.tsx
```diff
@@ -8,7 +8,7 @@ export default async function Page() {
   const { user } = await requireAuth();
 
   return (
-    <DefaultPageLayout>
+    <DefaultPageLayout layoutConfig={{ layout: 'default' }}>
```
rlaerm: Why is layout: 'default' used here? If default returns null, nothing is rendered here.

apps/dialog/src/components/layout/default-page-layout-client.tsx
rlaerm, line 1: File name should be changed to describe what this component is doing. Maybe PageLayoutSelector or something like this.

apps/dialog/src/components/layout/default-page-layout-client.tsx
```diff
@@ -15,70 +15,82 @@ import {
...
+function FormPageHeader() {
   const { formStateProps } = useCustomChatHeaderContent();
 
-  if (!header?.headerType) {
-    return null;
-  }
+  return (
+    <DialogHeaderContent>
+      <CustomChatHeader
+        showFormState={Boolean(formStateProps)}
```
rlaerm: I would prefer a standard nullish check `showFormState={formStateProps != null}`

apps/dialog/src/components/layout/default-page-layout-client.tsx
```diff
@@ -15,70 +15,82 @@ import {
...
-  if (header.headerType === 'form') {
-    return (
-      <DialogHeaderContent>
-        <CustomChatHeader
-          showFormState={Boolean(formStateProps)}
-          formStateProps={formStateProps ?? undefined}
-        />
-      </DialogHeaderContent>
-    );
+function DefaultPageHeader({ header }: { header?: DefaultPageLayoutConfig }) {
```
rlaerm: variable `header` should be renamed to `layoutConfig` or something similar.

apps/dialog/src/components/layout/default-page-layout-client.tsx
```diff
@@ -15,70 +15,82 @@ import {
...
+  if (!header) {
```
rlaerm: Why is parameter header optional? I would try to avoid optional parameters wherever it is possible.

apps/dialog/src/components/layout/default-page-layout-client.tsx
```diff
@@ -15,70 +15,82 @@ import {
...
-    return (
-      <DialogHeaderContent>
-        <div className="flex w-full gap-4">
-          <SelectImageModel />
-          <SelectImageStyle />
-        </div>
-      </DialogHeaderContent>
-    );
+  switch (header.layout) {
+    case 'default':
```
rlaerm: If we do not make use of `default`, I would remove it.

apps/dialog/src/components/layout/default-page-layout-client.tsx
```diff
@@ -15,70 +15,82 @@ import {
...
+      return null;
+    case 'form':
+      return <FormPageHeader />;
+    case 'image':
+      return (
+        <DialogHeaderContent>
```
rlaerm: case form returns FormPageHeader what is great. I would expect something similar for image and chat like ImagePageHeader and ChatPageHeader

recursive-hub (approved):

PR #904, opened 2026-05-11
fix: use primary color for svg logos

TobiasWallura-xitaso:


TobiasWallura-xitaso:
@copilot apply changes based on the comments in [this thread](https://github.com/FWU-DE/ais-chat/pull/904#pullrequestreview-4265237782)

apps/dialog/src/assets/logo-only.svg
rlaerm, line 1: I would suggest to use `fill="currentColor"`. In this case we can define the color that should be used for the svg via tailwind utility classes like 'text-primary' or 'text-red-500'.

rlaerm (approved):

apps/dialog/src/assets/logo-only.svg
TobiasWallura-xitaso: ok, will change it. the behavior will then be similar to phosphor icons: by default the icon is black and the color is defined by css classes (e.g. `text-primary`).

PR #945, opened 2026-05-18
Automatically copy model tables

AsamMax:


apps/admin/src/services/model-api-key-mapping-service.ts
```diff
@@ -27,5 +28,7 @@ export async function saveModelApiKeyMappings(
 
   logInfo('API Key mapping was updated successfully', { projectId, apiKeyId, modelIds });
 
+  await refreshAllModelsAfterSave();
```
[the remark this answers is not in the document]
AsamMax: fixed by speeding up

apps/admin/src/services/llm-service.ts
```diff
@@ -64,6 +67,8 @@ export async function updateLargeLanguageModel(
 
   logInfo('LLM was updated successfully', { organizationId, modelId, data });
 
+  await refreshAllModelsAfterSave();
+
   if (!model) throw new Error('Failed to update model');
```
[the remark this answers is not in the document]
AsamMax: fixed by speeding up

apps/admin/src/services/llm-service.ts
```diff
@@ -36,6 +37,8 @@ export async function createLargeLanguageModel(
 
   logInfo('LLM was created successfully', { organizationId, data });
 
+  await refreshAllModelsAfterSave();
+
   if (!model) throw new Error('Failed to create model');
```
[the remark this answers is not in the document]
AsamMax: fixed by speeding up

apps/admin/src/services/model-refresh-service.ts
```diff
@@ -0,0 +1,11 @@
+import { dbUpdateLlmModelsForAllFederalStates } from '@shared/db/functions/llm-model';
+import { logError } from '@shared/logging';
+
+export async function refreshAllModelsAfterSave() {
+  try {
+    await dbUpdateLlmModelsForAllFederalStates();
+  } catch (error) {
+    logError('Error refreshing LLM models after save', error);
+    throw error;
```
TobiasWallura-xitaso: Log and rethrow should usually not be done. IMO we don't need to log here.

apps/admin/src/services/llm-service.ts
```diff
@@ -36,6 +37,8 @@ export async function createLargeLanguageModel(
 
   logInfo('LLM was created successfully', { organizationId, data });
 
+  await refreshAllModelsAfterSave();
```
TobiasWallura-xitaso: Why do we need the refresh here? in this fn we just add a new llm, without linking it to any api key.

apps/admin/src/services/model-refresh-service.ts
```diff
@@ -0,0 +1,11 @@
+import { dbUpdateLlmModelsForAllFederalStates } from '@shared/db/functions/llm-model';
+import { logError } from '@shared/logging';
+
+export async function refreshAllModelsAfterSave() {
```
TobiasWallura-xitaso: Do we really need to refresh all models? Won't it be more efficient to refresh only affected federal states / models?

packages/shared/src/db/functions/llm-model.ts
```diff
@@ -152,12 +170,24 @@ export async function dbUpsertLlmModelsByModelsAndFederalStateId({
...
+}: {
+  federalStateId: string;
+  modelIds: string[];
+}) {
+  for (const modelId of modelIds) {
     await db
       .insert(federalStateLlmModelMappingTable)
-      .values({ federalStateId, llmModelId: model.id })
+      .values({ federalStateId, llmModelId: modelId })
       .onConflictDoNothing();
```
TobiasWallura-xitaso: Prefer to use a single SQL query instead of for loop (e.g. provide an array to `.values(...)`

TobiasWallura-xitaso (approved):

apps/admin/src/services/model-refresh-service.ts
AsamMax: It would be more efficient, but it would add a bunch of complexity. Also: dbUpdateLlmModelsForAllFederalStates checks before it updates, so it updates as little as reasonably possible.

brueckb (approved):

PR #948, opened 2026-05-19
fix loading spinner in downloadButton for dev setup

brueckb:


brueckb:
this is more complex than it looks at first glance: The introduced changes in this PR led to the file being downloaded twice when triggered via the three dots menu in mobile. Since the endless loading spinner that was to be fixed here is a problem we only encounter in dev setup due to react strict mode, we decided to drop the topic for now.

PR #950, opened 2026-05-20
chore: update dependencies to fix vulnerabilities

TobiasWallura-xitaso:


package.json
```diff
@@ -32,7 +32,10 @@
     "overrides": {
       "glob@<11.1.0": ">=11.1.0",
       "minimatch@~3.1.4": "^3.1.4",
-      "minimatch@~9.0.0": "^9.0.7"
+      "minimatch@~9.0.0": "^9.0.7",
```
rlaerm: I would like to discuss our strategy here. I am not sure how to handle that and how to keep an eye on that.

rlaerm (approved):

brueckb (approved):

package.json
TobiasWallura-xitaso: We decided to target only the specific affected ranges and update to the minimum fixed/unaffected version.

PR #956, opened 2026-05-21
feat/TD-699: database migration and service layer for reports

brueckb:


packages/shared/src/reports/report-service.ts
```diff
@@ -0,0 +1,301 @@
...
+
+export async function getEntityReportOverviews({
+  limit,
+}: {
+  limit?: number;
+} = {}): Promise<EntityReportOverview[]> {
+  const allReports = await dbGetAllEntityReports();
+
+  const groupedReports = new Map<string, Awaited<ReturnType<typeof dbGetAllEntityReports>>>();
+  for (const report of allReports) {
```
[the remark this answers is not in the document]
brueckb: validation of the limit has been moved before the db read. I decided to stick to the full db read for now and address it if we see that it is needed based on the number of reports we receive.

packages/shared/src/db/schema.ts
```diff
@@ -392,6 +407,7 @@ export const characterTable = pgTable(
       .array()
       .notNull()
       .default(sql`'{}'::text[]`),
+    suspended: boolean('suspended').notNull().default(false),
```
TobiasWallura-xitaso, line 413: The `suspended` must be updatable only by an admin. We must ensure that the user cannot update it (e.g. by modifying an HTTP request). -> should probably be included in `omit` by the insert/update zod schemas.

packages/shared/src/suspension/suspension-service.ts
```diff
@@ -0,0 +1,343 @@
...
+    },
+  });
+}
+
+export async function markReportAsChecked(reportId: string) {
+  checkParameterUUID(reportId);
+  return dbMarkEntityReportAsChecked({ reportId });
+}
+
+export async function suspendEntity({
```
TobiasWallura-xitaso, line 176: Should we merge `suspendEntity` and `liftSuspensionOnEntity`?

packages/shared/src/suspension/suspension-service.ts
```diff
@@ -0,0 +1,343 @@
...
+}
+
+export async function getEntityReportOverviews({
+  limit,
+}: {
+  limit?: number;
+} = {}): Promise<EntityReportOverview[]> {
+  if (limit !== undefined && (!Number.isInteger(limit) || limit < 1)) {
+    throw new InvalidArgumentError('limit must be a positive integer');
+  }
```
TobiasWallura-xitaso: Do we need the `limit` or can it all be removed?

packages/shared/src/db/functions/assistants.ts
TobiasWallura-xitaso: This can be deleted.

stschaller-xitaso (approved):

TobiasWallura-xitaso (approved):

packages/shared/migrations/0072_cold_taskmaster.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TYPE "public"."suspension_request_reason" AS ENUM('copyright_violation', 'false_or_outdated_information', 'insufficient_sources', 'discrimination', 'personal_data_usage_or_query', 'violence_or_extremist_content', 'sexualized_content', 'other');--> statement-breakpoint
```
rlaerm, line 1: Personally I would not define that on database layer. Every time a new value is added or an existing one is removed, you have to write a database migration.

packages/shared/src/db/functions/character.ts
```diff
@@ -458,6 +471,31 @@ export async function dbGetGlobalCharacterByName({
...
+  suspended: boolean;
+}) {
+  const [updatedCharacter] = await db
+    .update(characterTable)
+    .set({ suspended })
+    .where(eq(characterTable.id, characterId))
+    .returning();
+
+  if (!updatedCharacter) {
+    throw new NotFoundError('Character not found');
```
rlaerm, line 488: That works differently than most of the other db functions. It is fine to me but we have to check now if a db function throws an error or return undefined.

rlaerm (approved):

PR #992, opened 2026-06-05
use shared ui buttons for login and disclaimer

brueckb:


apps/chat-bot/src/components/modals/terms-conditions-initial.tsx
```diff
@@ -82,26 +82,26 @@ export default function TermsConditionsModal({
...
         >
           {tCommon('accept')}
-        </button>
+        </Button>
       ) : (
-        <button
+        <Button
           onClick={nextPage}
-          className={buttonPrimaryClassName}
+          className="py-3 h-auto"
```
stschaller-xitaso: Would it make sense to add this style (py-3 h-auto) to a variant or custom Style since it is used multiple times including the other files?

stschaller-xitaso (approved):

PR #1001, opened 2026-06-08
TD-1169: add correct touch behavior for mobile in image upload chat

stschaller-xitaso:


recursive-hub (approved):

apps/chat-bot/src/components/custom-chat/custom-chat-avatar-crop-modal.tsx
```diff
@@ -139,6 +142,7 @@ export default function AvatarCropModal({
                   }
                 }}
                 className="h-auto w-auto max-w-none"
+                style={{ touchAction: 'pan-y' }}
```
TobiasWallura-xitaso, line 147: Can we use the tailwind class `touch-pan-y` here as well?
```suggestion
                className="h-auto w-auto max-w-none touch-pan-y"
```

TobiasWallura-xitaso (approved):

brueckb (approved):

apps/chat-bot/src/components/custom-chat/custom-chat-avatar-crop-modal.tsx
stschaller-xitaso: The ReactCrop.css sets explicitly tocuh-action: none for the img, so in order to have it in the classname and still override the ReactCrop.css I would have to set !important, which is also not great. I would leave as is and add a comment

PR #1003, opened 2026-06-09
Support Ionos in Agentic Mode + Multiple Tool calls

AsamMax:


packages/ai-core/src/chat/providers/ionos.ts
```diff
@@ -118,34 +118,113 @@ export function constructIonosAgenticStreamFn(model: AiModel): AgenticStreamFn {
...
+      usage = {
+        completionTokens: calculatedUsage.completion_tokens,
+        promptTokens: calculatedUsage.prompt_tokens,
+        totalTokens: calculatedUsage.total_tokens,
+      };
+    }
+
+    if (!usage) {
+      throw new AiGenerationError('No usage data returned from IONOS stream');
+    }
```
TobiasWallura-xitaso: This is dead code.

packages/ai-core/src/chat/providers/ionos.ts
```diff
@@ -118,34 +118,113 @@ export function constructIonosAgenticStreamFn(model: AiModel): AgenticStreamFn {
...
+
+        toolCalls.set(toolCallDelta.index, existingToolCall);
+      }
+    }
+
+    const resolvedToolCalls: ToolCall[] = [...toolCalls.entries()]
+      .sort(([left], [right]) => left - right)
+      .map(([, toolCall]) => {
+        if (!toolCall.id || !toolCall.name) {
+          throw new AiGenerationError('Incomplete tool call returned from IONOS stream');
```
TobiasWallura-xitaso: Throwing will prevent the costs to be tracked, right?

packages/ai-core/src/chat/providers/ionos.ts
```diff
@@ -118,34 +118,113 @@ export function constructIonosAgenticStreamFn(model: AiModel): AgenticStreamFn {
...
-    });
+      });
+
+    if (!usage) {
+      const completionContent = [
+        content,
+        ...resolvedToolCalls.map((toolCall) =>
+          JSON.stringify({ name: toolCall.name, arguments: toolCall.arguments }),
+        ),
+      ].join('');
```
TobiasWallura-xitaso, line 201: For understanding: IONOS (sometimes) does not include the usage/tokens in the response. Therefore we calculate an approximation.
Instead of concatenating the content and tool calls, can't we just use the whole `stream`? The `usage` anyways is not part of it, inside this `if` case.

TobiasWallura-xitaso (approved):

packages/ai-core/src/chat/providers/ionos.ts
AsamMax: But the entire stream also contains information that isn't generated by the LLM and as such isn't billed usage, right? Or am i misunderstanding your comment?

PR #1007, opened 2026-06-09
TD 1193 Agentic Custom Chats

AsamMax:


apps/chat-bot/src/app/api/chat/chat-service.ts
```diff
@@ -263,21 +245,57 @@ export async function sendChatMessage({
...
+    | undefined;
+  let activeToolDefinitions: ToolDefinition[] = [];
+
+  const urls = extractUrls({
+    assistant: activeAssistant,
+    character: activeCharacter,
+    learningScenario: activeLearningScenario,
+    messages,
+  });
+  const ingestResult = await ingestWebContent({
```
recursive-hub, line 265: Let's discuss whether in the regular chat, the urls from the user messages should be automatically extracted and ingested.

apps/chat-bot/e2e/tests/api/v1/admin/cleanup/cleanup.api.test.ts
```diff
@@ -164,7 +164,7 @@ test.describe('cleanup', () => {
     expect(resultExisting).toHaveLength(1);
   });
 
-  test('should delete web chunks older than 30 days', async ({ request }) => {
+  test('should keep web chunks during cleanup', async ({ request }) => {
```
recursive-hub: another option would be to ingest web content chunks if the web scraper tool is called and if the chunks are not already in the vector db. Then it would be the same logic as right now.

recursive-hub (approved):

PR #1014, opened 2026-06-11
Feature:Download-individual-messages

sammo3332:


apps/chat-bot/src/app/api/chat/chat-service.ts
```diff
@@ -283,6 +283,16 @@ export async function sendChatMessage({
...
+    await dbInsertChatContent({
+      id: assistantMessageId,
+      content: fullText,
+      role: 'assistant',
+      userId: user.id,
+      orderNumber: assistantMessageOrderNumber,
+      modelName: definedModel.name,
+      conversationId: activeConversation.id,
+      webSearchResults,
+    });
```
TobiasWallura-xitaso: Looks like a merge conflict, this file should not need any changes.
```suggestion
```

packages/shared/src/db/functions/chat.ts
```diff
@@ -68,6 +68,31 @@ export async function dbGetConversationMessages({
...
+  userId,
+  conversationId,
+  messageId,
+}: {
+  userId: string;
+  conversationId: string;
+  messageId: string;
+}) {
+  const [row] = await db
+    .select()
```
TobiasWallura-xitaso: minor for performance reasons: select only `conversation_message` as the conversation is not used further on.

apps/chat-bot/src/app/api/download-conversation/utils.ts
```diff
@@ -111,6 +111,96 @@ function getConversationMessages({
   ]);
 }
 
+export async function generateConversationMessageDocxFile({
```
TobiasWallura-xitaso, line 114: Can we reuse the existing function `generateConversationDocxFile` and pass only the single message?
There are only slight differences in the metadata paragraphs; maybe they should be the same.

sammo3332: I briefly reused generateConversationDoxFile, but that also brought back what Elena would, Elena aligned a different layout for single-message exports, I kept the message export separate. Shared parts like markdownToDocx, buildDocxDocument, and model displayname are still reused.

packages/shared/src/db/functions/chat.ts
sammo3332: Adjusted

TobiasWallura-xitaso (approved):

PR #1027, opened 2026-06-17
Feat/td 1319 claude sonnet

rlaerm:


apps/chat-bot/src/app/api/chat/utils.ts
```diff
@@ -198,7 +208,20 @@ export async function getChatTitle({
 }
 
 /**
- * Converts frontend messages to ai-core message format
+ * Some models (like google anthropic) require the image data to be included in the message as a base64 encoded string,
+ * while others can work with just the image url. This function conditionally includes the base64 encoded data if required by the model.
+ */
+export function determineImageAttachmentTypeForModel(model: LlmModelSelectModel): 'url' | 'base64' {
+  // we do not have settings on the LlmModelSelectModel to determine if the model needs image data,
+  // so we will use the model name as a heuristic for now
```
brueckb, line 216: what do you think about storing this in the db, e.g. as supported_image_attachment_type(s)?

apps/chat-bot/src/app/api/rag/rag-service.ts
```diff
@@ -70,7 +77,10 @@ export async function retrieveChunks({
   relatedFileEntities: FileModelAndContent[];
   sourceUrls?: string[];
 }): Promise<RetrievedChunk[]> {
-  if (relatedFileEntities.length === 0 && (!sourceUrls || sourceUrls.length === 0)) {
+  // Todo RL. Filter out images
```
brueckb: is this an oversight or on purpose?

brueckb (approved):

apps/chat-bot/src/app/api/chat/utils.ts
rlaerm: In general, yes. We already have configuration for providers. But that is part of apps/api and the configuration is not taken over from api to chat-bot. Second, we want to introduce a package that handles the abstraction for us in the near future, so every effort in this direction may be obsolete.

apps/chat-bot/src/app/api/rag/rag-service.ts
rlaerm: No, I just forgot to remove the comment.

AsamMax (approved):

PR #1032, opened 2026-06-17
TD-1288: fix SessionWatcher to prevent logout on network failures

TobiasWallura-xitaso:
This pull request introduces a new `ResilientSessionProvider` to prevent unwanted user logouts in the chat-bot app when the session endpoint is unavailable due to network or infrastructure errors. It ensures that the application uses a cached session during transient failures, only logging out users on genuine session invalidation. The PR also adds comprehensive end-to-end tests to verify this resilience.

### DIFF
--- a/apps/chat-bot/src/auth/ResilientSessionProvider.tsx
+++ b/apps/chat-bot/src/auth/ResilientSessionProvider.tsx
@@ -28,6 +28,7 @@ export default function ResilientSessionProvider({
 }: ResilientSessionProviderProps) {
   const lastValidSessionRef = useRef<Session | null>(session);
   const originalFetchRef = useRef<typeof fetch | null>(null);
+  const wrappedFetchRef = useRef<typeof fetch | null>(null);
 
   useEffect(() => {
     // Store original fetch if not already stored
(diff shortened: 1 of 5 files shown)


apps/chat-bot/src/auth/ResilientSessionProvider.tsx
```diff
@@ -0,0 +1,135 @@
...
+              headers: { 'Content-Type': 'application/json' },
+            });
+          }
+        }
+
+        // 4xx or no cached session - allow the error to propagate
+        return response;
+      } catch (error) {
+        // Network error (fetch failed)
+        logError('Session endpoint unreachable - using cached session', error);
```
rlaerm: Is that still an error now? Or should we go with warning instead?

apps/chat-bot/src/auth/ResilientSessionProvider.tsx
```diff
@@ -0,0 +1,135 @@
...
+    if (!originalFetch) {
+      throw new Error('ResilientSessionProvider: original fetch not available');
+    }
+
+    // Intercept fetch for session endpoint
+    const wrappedFetch = async (input: RequestInfo | URL, init?: RequestInit) => {
+      const url = typeof input === 'string' ? input : input instanceof URL ? input.href : input.url;
+
+      // Only intercept session endpoint requests (exact match on pathname)
+      const SESSION_ENDPOINT = '/api/auth/session';
```
rlaerm: I would move the const to line 7.

rlaerm (approved):

PR #1033, opened 2026-06-17
Feat/TD-1211: filtering of entities

stschaller-xitaso:


packages/shared/src/db/schema.ts
```diff
@@ -382,6 +382,128 @@ export const suspensionRequestReasonEnum = pgEnum(
...
+  'vocational-school',
+  'technical-college',
+  'other',
+]);
+export type SchoolType = z.infer<typeof schoolTypesSchema>;
+
+export const gradeRangesSchema = z.enum(['range-1', 'range-2', 'range-3', 'range-4']);
+export type GradeRange = z.infer<typeof gradeRangesSchema>;
+
+export const categoriesSchema = z.enum([
```
brueckb, line 402: should we consider just using strings for some of the filter categories? I could imagine them changing a lot, especially in the beginning. that would mean a lot of migrations...

brueckb (approved):

packages/shared/src/db/schema.ts
stschaller-xitaso: I have considered that as well, but in this case I still chose enums, because we use Zod enums here, not postrges enums, meaning any changes in the filter options would not require db migrations directly (in the migrations in this PR only a quite generic column "filter-attributes" filter is added and old redundant filters are removed).

But if filter options change, translation Keys have to be adapted and old JSONs may not pass validation, but then with strings we would also loose Zod enum validation completely...
You're right that a "hybrid option" might be good, so "stable" options which are not likely to change soon (federal states, etc) are still enums and others are strings.
Or maybe adding an "adapter" for renamed values, so we do not have errors, if there are outdated values 🤔

stschaller-xitaso: Happy to discuss it next week 😊

apps/chat-bot/src/components/custom-chat/custom-chat-filter/custom-chat-filter-utils.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export type ActiveFilterPill = {
+  label: string;
+  group: keyof FilterValues;
+  value: string;
+};
+
+export function getActiveFilterPills(
+  values: FilterValues,
+  t: (key: string) => string,
```
TobiasWallura-xitaso, line 126: Try to pass only `t` into this function (instead of `(key) => t(key)`) and use this type for type-safety and improved code completion.
```suggestion
  t: ReturnType<typeof useTranslations>,
```

Same applies to other functions as well.

packages/shared/src/db/schema.ts
```diff
@@ -382,6 +382,128 @@ export const suspensionRequestReasonEnum = pgEnum(
...
+  'french',
+  'italian',
+  'spanish',
+  'greek',
+  'latin',
+  'russian',
+]);
+export type Language = z.infer<typeof languagesSchema>;
+
+export const subjectsSchema = z.enum([
```
TobiasWallura-xitaso: This enum could also be split into multiple enums for each sub-category (and if needed, combine all of the sub-category enums into a single one again, e.g. `const mergedEnum = z.enum([...a.options, ...b.options] as const)`).
Then we would not need to duplicate the enum values in `custom-chat-filter-keys.ts`.

apps/chat-bot/src/components/custom-chat/custom-chat-filter/custom-chat-filter-utils.ts
```diff
@@ -0,0 +1,167 @@
...
+    federal_states?: string[];
+    languages?: string[];
+  };
+  schoolType?: string | null;
+  gradeLevel?: string | null;
+  subject?: string | null;
+};
+
+function unique(values: string[]): string[] {
+  return values.filter((value, index, allValues) => allValues.indexOf(value) === index);
```
TobiasWallura-xitaso: In other places we use `...new Set(values)` (more performant than filter + search)

TobiasWallura-xitaso (approved):

apps/chat-bot/src/components/custom-chat/custom-chat-filter/custom-chat-filter-utils.ts
stschaller-xitaso: If I do that in this function I have a linter error later in the return type and have to write:
`label: t(`school-types.${value}` as unknown as Parameters<typeof t>[0]),`
which negates the benefit of using ReturnType<typeof useTranslations> so I chose to keep `t: (key: string) => string`.
If you have another idea how to prevent the linter errors, please let me know 😊

stschaller-xitaso: Did however adapt it in `UseOverviewFiltersOptions`

stschaller-xitaso: Kommando zurück, da bekomme ich den gleichen Linterfehler mit gleicher "Lösung" -> habs jetzt mal reverted

TobiasWallura-xitaso: Check what Claude did in this PR: https://github.com/FWU-DE/ais-chat/pull/1051

PR #1050, opened 2026-06-22
TD-1326: soft delete regenerated msg and clean up db

stschaller-xitaso:


packages/shared/migrations/0079_delete_duplicate_assistant_messages.sql
```diff
@@ -0,0 +1,19 @@
...
+		ROW_NUMBER() OVER (
+			PARTITION BY conversation_id, order_number, role
+			ORDER BY created_at DESC, id DESC
+		) AS duplicate_rank
+	FROM conversation_message
+	WHERE role = 'assistant'
+		AND deleted_at IS NULL
+)
+UPDATE conversation_message
+SET content = ' ',
```
brueckb, line 13: are we 100% sure we want to delete the content? If we ignore messages with non-empty `deleted_at`, we don't need to do that if we see any risk in that. (same for dbDeleteRegeneratedConversationMessage)

brueckb (approved):

packages/shared/migrations/0079_delete_duplicate_assistant_messages.sql
stschaller-xitaso: Discussed with Tobi and Basti and would bi a bigger scope than this PR as it is questioning the point of soft deleting in general -> has been added to [technical debts](https://fwu-de.atlassian.net/wiki/spaces/KIChatbot/pages/670564354/11.2+-+Technische+Schulden)

PR #1054, opened 2026-06-23
Fix agentic e2e failures

AsamMax:
## JIRA Ticket

TD-1193

## Bug Description

Two external-services e2e cases failed under agentic chat: GPT-4o-mini returned a 404 via Azure Responses API, and the web-search test never received web search sources.

- **Azure agentic streaming**
  - Route non-GPT-5 Azure models through chat completions for agentic streams.
  - Keep GPT-5-family models on Azure Responses API.
  - Validate streamed tool-call arguments before yielding them.

- **Mock LLM web search path**
  - Emit deterministic `web_search` tool calls from `/v1/responses` for current-info prompts.
  - Avoid repeated web-search calls once a function-call output exists.

## How to Reproduce (Original Bug)

Run the external-services e2e suite with agentic chat enabled. The failures appear in:

- `all-models.test.ts` for `GPT-4o-mini responds to a simple prompt`
- `web-search.test.ts` for `should successfully perform web search`

## Screenshots (if appropriate)

N/A

### DIFF
--- a/apps/chat-bot/e2e/tests/assistant/assistant-file-upload.test.ts
+++ b/apps/chat-bot/e2e/tests/assistant/assistant-file-upload.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from '@playwright/test';
-import { AUTH_FILES, MOCK_LLM_COMMANDS } from '../../utils/const';
+import { AUTH_FILES } from '../../utils/const';
 import { sendMessage, uploadFile } from '../../utils/chat';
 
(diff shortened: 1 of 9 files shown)


AsamMax:
@copilot Please take a look at the failing e2e tests. The new agentic mode probably breakes some tests, because now tools aren't automatically called before. For every failing test take a moment to reflect if it still makes sense to have, or should be discarded. If we keep it, please fix it. Prefer clean solutions over complex ones. Do not change other functionality like image generation

PR #1063, opened 2026-06-24
add instrumentation of anthropic models

rlaerm:


packages/ai-core/src/chat/providers/google-anthropic.ts
```diff
@@ -196,6 +197,13 @@ export function constructGoogleAnthropicAgenticStreamFn(model: AiModel): Agentic
   };
 }
 
+function createAnthropicClient(options: ClientOptions): AnthropicVertex {
+  return instrumentAnthropicAiClient(new AnthropicVertex(options), {
+    recordInputs: true,
+    recordOutputs: true,
+  });
```
TobiasWallura-xitaso, line 204: We should not record inputs/outputs for privacy reasons.
```suggestion
  return instrumentAnthropicAiClient(new AnthropicVertex(options));
```

rlaerm: Yes, I will remove it as soon as it is working.

TobiasWallura-xitaso (approved):

PR #1078, opened 2026-06-26
TD-1345: add tests and logs for chat messages

stschaller-xitaso:


packages/shared/src/db/functions/chat.ts
```diff
@@ -3,6 +3,63 @@ import { db } from '..';
...
+  type LogDataType = Record<string, unknown>;
+  const logData: LogDataType = {
+    totalSkipped,
+  };
+
+  if (skippedMessages.length > 0) {
+    logData.skippedMessages = skippedMessages;
+  }
+
+  logWarning('Skipped conversation message batch inserts due to conflict.', logData);
```
rlaerm: I suggest to define a new function like logAssertion() that logs an error instead of warning. That should not happen and if it happens we should notice that as soon as possible.

packages/shared/src/db/functions/chat.ts
```diff
@@ -3,6 +3,63 @@ import { db } from '..';
 import { conversationMessageTable, conversationTable } from '../schema';
 import type { ConversationMessageModel, InsertConversationMessageModel } from '../types';
 import { isNotNull } from '../../utils/guard';
+import { logWarning } from '../../logging/logging';
+
+function getConflictLogData(chatContent: InsertConversationMessageModel) {
```
rlaerm: Is that really necessary? It is exactly the same model with one renaming.

packages/shared/src/db/functions/chat.ts
```diff
@@ -3,6 +3,63 @@ import { db } from '..';
...
+  return {
+    conversationId: chatContent.conversationId,
+    messageId: chatContent.id,
+    orderNumber: chatContent.orderNumber,
+    role: chatContent.role,
+    userId: chatContent.userId,
+  };
+}
+
+function logSingleInsertConflict(chatContent: InsertConversationMessageModel) {
```
rlaerm: I think you can get rid of this function and use the logBatchInsertConflicts function instead

packages/shared/src/db/functions/chat.ts
```diff
@@ -90,13 +147,35 @@ export async function dbGetConversationMessageById({
...
-    await db.insert(conversationMessageTable).values(chatContent).onConflictDoNothing().returning()
-  )[0];
+  const [insertedMessage] = await db
+    .insert(conversationMessageTable)
+    .values(chatContent)
+    .onConflictDoNothing()
+    .returning();
+
+  if (!insertedMessage) {
+    logSingleInsertConflict(chatContent);
```
rlaerm: use the logBatchInsertConflicts() function here and pass insertedMessage as well

rlaerm (approved):

packages/shared/src/db/functions/chat.ts
stschaller-xitaso: I used the logError method from the logger

PR #1083, opened 2026-06-30
fix streaming of short messages

rlaerm:


packages/ai-core/src/chat/providers/google-anthropic.ts
```diff
@@ -346,6 +356,23 @@ function ensureValidStopReason(reason: StopReason | null) {
...
+ * Checks if a tool result string represents an error.
+ * Tries to parse it as JSON and checks for an "error" field with a non-null value.
+ */
+function isToolResultError(content: string): boolean {
+  try {
+    if (content.startsWith('Error:')) return true;
+
+    const parsed = JSON.parse(content);
+    // Check if the parsed object has an "error" field with a truthy value
+    return parsed && typeof parsed === 'object' && 'error' in parsed && parsed.error !== null;
```
TobiasWallura-xitaso, line 369: The 'error' key is from the `xxxToolResult` types, right? Can we link the parsing here to the typings somehow?

packages/ai-core/src/chat/providers/google-anthropic.ts
```diff
@@ -393,10 +420,17 @@ function mapMessageToAnthropicMessageParam(message: NonSystemMessage): MessagePa
...
+
+    // Check if this is an error response (either starts with "Error:" or contains error in JSON)
+    const isErrorResult = isToolResultError(toolResultContent);
+
     content.push({
       type: 'tool_result',
       tool_use_id: message.toolCallId ?? '',
-      content: message.content,
+      content: toolResultContent,
+      is_error: isErrorResult,
```
TobiasWallura-xitaso, line 433: Should we add a test case for `is_error: true`?

packages/ai-core/src/chat/agent-loop.test.ts
```diff
@@ -0,0 +1,337 @@
...
+      toolRegistry,
+      onTextChunk,
+      onComplete,
+      onError,
+    });
+
+    // Wait for async completion
+    await vi.waitFor(() => {
+      expect(onComplete).toHaveBeenCalled();
+    });
```
TobiasWallura-xitaso, line 76: what's this? there should be an async expect

packages/ai-core/src/chat/providers/google-anthropic.ts
```diff
@@ -346,6 +356,23 @@ function ensureValidStopReason(reason: StopReason | null) {
     throw new AiGenerationError('Model refused to generate a response during agentic generation');
 }
 
+/**
+ * Checks if a tool result string represents an error.
+ * Tries to parse it as JSON and checks for an "error" field with a non-null value.
+ */
+function isToolResultError(content: string): boolean {
+  try {
+    if (content.startsWith('Error:')) return true;
```
TobiasWallura-xitaso, line 365: Should we instead change all tools to return stringified jsons? Two different solutions for how errors are returned looks like tech debt to me.

TobiasWallura-xitaso (approved):

packages/ai-core/src/chat/agent-loop.test.ts
rlaerm: Sorry, but I don't know what you mean and how that should look like. Can you explain that in more detail?

TobiasWallura-xitaso: https://main.vitest.dev/guide/recipes/wait-for

packages/ai-core/src/chat/providers/google-anthropic.ts
rlaerm: We created a tech debt item for typings and error handling in tool calling.

packages/ai-core/src/chat/providers/google-anthropic.ts
rlaerm: The typing is bad. We get a string which is a stringified object.

packages/ai-core/src/chat/providers/google-anthropic.ts
rlaerm: Done

PR #1088, opened 2026-06-30
Feat/td 310 share with learners add time and raise token limit

brueckb:


packages/shared/src/db/functions/learning-scenario.ts
```diff
@@ -422,6 +422,80 @@ export function dbGetSharedLearningScenarioConversations({
...
+  const [latestUnstoppedShare] = await db
+    .select()
+    .from(sharedLearningScenarioTable)
+    .where(
+      and(
+        eq(sharedLearningScenarioTable.learningScenarioId, learningScenarioId),
+        eq(sharedLearningScenarioTable.userId, user.id),
+        isNull(sharedLearningScenarioTable.manuallyStoppedAt),
+      ),
+    )
```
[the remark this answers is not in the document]
brueckb: we do want to be able to do that for expired chats, since they should be extendable for a certain amount of time after expiredAt. Might be worth considering if we should put that amount of time into the query.

apps/chat-bot/src/app/(authed)/(chat-bot)/learning-scenarios/_components/count-down.tsx
```diff
@@ -20,18 +20,28 @@ export default function CountDownTimer({
...
   const [timeRemaining, setTimeRemaining] = React.useState(Math.max(leftTimeInSeconds, 0));
 
+  // synchronize internal state with changes to leftTimeInSeconds (i.e. extend usage time)
+  React.useEffect(() => {
+    const resetTimer = window.setTimeout(() => {
+      setTimeRemaining(Math.max(leftTimeInSeconds, 0));
+    }, 0);
+
+    return () => window.clearTimeout(resetTimer);
+  }, [leftTimeInSeconds]);
```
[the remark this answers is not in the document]
brueckb: Calling setState synchronously within an effect body causes cascading renders that can hurt performance, and is not recommended. (https://react.dev/learn/you-might-not-need-an-effect).

apps/chat-bot/src/components/custom-chat/custom-chat-share-with-learners/custom-chat-adjust-token-limit-button.tsx
```diff
@@ -0,0 +1,135 @@
...
+  currentTokenPointsPercentageLimit: number;
+  maxAvailablePercentage: number;
+  onAdjustTokenLimit: (data: { tokenPointsPercentageLimit: number }) => Promise<{
+    success: boolean;
+    tokenPointsLimit?: number;
+  }>;
+  onAdjustTokenLimitSuccess: (newTokenPointsLimit: number) => void;
+};
+
+export function CustomChatAdjustTokenLimitButton({
```
rlaerm: I would suggest to create a custom-chat-adjust-token-limit-dialog component instead of the button. IMHO that is more intuitive and that is what I would search for. The button itself is just a small component that triggers the dialog but can also be passed as prop. The main functionality is about the dialog, not the button. I know we have a lot of buttons that work the same way but I would prefer to have a dialog component. Do you agree?

brueckb: yes! will do :)

apps/chat-bot/src/components/custom-chat/custom-chat-share-with-learners/custom-chat-adjust-token-limit-button.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  async function handleAdjustTokenLimit() {
+    setIsSubmitting(true);
+    try {
+      const result = await onAdjustTokenLimit({ tokenPointsPercentageLimit });
+      if (result.success && result.tokenPointsLimit !== undefined) {
+        toast.success(tToast('adjust-token-limit-toast-success'));
+        setOpen(false);
+        onAdjustTokenLimitSuccess(result.tokenPointsLimit);
+        // early return runs through finally and avoids the error toast
```
rlaerm: Why do we not place the error toast in the catch block? I think it is easier to read/understand and we do not need the comment and the return statement.

apps/chat-bot/src/components/custom-chat/custom-chat-share-with-learners/custom-chat-adjust-token-limit-button.tsx
```diff
@@ -0,0 +1,135 @@
...
+        <DialogHeader className="space-y-2">
+          <DialogTitle>{t('adjust-token-limit-modal-title')}</DialogTitle>
+          <DialogDescription>{t('adjust-token-limit-modal-description')}</DialogDescription>
+        </DialogHeader>
+        <div>
+          <TokenPointsLimitSelect
+            label={t('token-points')}
+            ariaLabel={t('adjust-token-limit-modal-aria')}
+            defaultValue={String(nextHigherOption)}
+            value={String(tokenPointsPercentageLimit)}
```
rlaerm: Why is the value a string? onValueChange() receives a number, pointsPercentageValues is a number, everything seems to be a number in TokenPointsLimitSelect but the value is not.

apps/chat-bot/src/components/custom-chat/custom-chat-share-with-learners/custom-chat-extend-share-expiration-button.tsx
```diff
@@ -0,0 +1,117 @@
...
+        // early return runs through finally and avoids the error toast
+        return;
+      }
+    } catch {
+      // Errors are handled by the generic failure toast below.
+    } finally {
+      setIsSubmitting(false);
+    }
+
+    toast.error(tToast('add-time-toast-error'));
```
rlaerm: move toast in catch block plz

apps/chat-bot/src/components/custom-chat/custom-chat-share-with-learners/custom-chat-share-with-learners.tsx
```diff
@@ -138,25 +148,34 @@ export function CustomChatShareWithLearners({
                         <div className="pr-6">
                           <TokenPointsLimitSelect
                             ariaLabel={t('max-token-points')}
-                            defaultValue={String(getValuesShare('tokenPointsPercentageLimit'))}
+                            defaultValue={String(currentTokenPointsPercentageLimit)}
```
rlaerm: Why do we not use a number here?

packages/shared/src/characters/character-service.ts
```diff
@@ -487,6 +489,95 @@ export const unshareCharacter = async ({
   return updatedCharacter;
 };
 
+/**
+ * Extends the expiration of an active character share.
+ * @throws NotFoundError if no active sharing exists for the character.
```
rlaerm: I would prefer a InvalidArgumentError here. The entity is the character. So if there is no active sharing for that character it is an invalid operation atm which should result in an InvalidArgumentError.

packages/shared/src/characters/character-service.ts
```diff
@@ -487,6 +489,95 @@ export const unshareCharacter = async ({
...
+  if (!updatedShare) {
+    throw new NotFoundError('No active sharing found for this character');
+  }
+
+  return updatedShare;
+};
+
+/**
+ * Increases the token points limit of an active character share.
+ * @throws NotFoundError if no sharing exists for the character.
```
rlaerm: same as above

packages/shared/src/db/functions/character.ts
```diff
@@ -544,3 +544,87 @@ export async function dbGetSharedCharacterConversations({
...
+        eq(sharedCharacterConversation.userId, user.id),
+        isNull(sharedCharacterConversation.manuallyStoppedAt),
+        sql`${sharedCharacterConversation.expiredAt} >= now()`,
+      ),
+    )
+    .orderBy(desc(sharedCharacterConversation.startedAt))
+    .limit(1);
+
+  if (!latestUnstoppedShare) {
+    return undefined;
```
rlaerm, line 576: I think returning null is the correct value here. Undefined has the meaning of "no value provided", especially for optional paramters but also for return values. null has the meaning of "we looked for this value, but did not found anything" or for parameters "we explicitely pass an empty value here". I know it is not consistent in our codebase but that is how I would handle it.

packages/shared/src/db/functions/character.ts
```diff
@@ -544,3 +544,87 @@ export async function dbGetSharedCharacterConversations({
...
+  characterId: string;
+  user: Pick<UserModel, 'id'>;
+  additionalTimeInMinutes: number;
+}) {
+  const [latestUnstoppedShare] = await db
+    .select()
+    .from(sharedCharacterConversation)
+    .where(
+      and(
+        eq(sharedCharacterConversation.characterId, characterId),
```
rlaerm, line 566: Does it make sense to create a new index on characterId? I am not sure if we do any kind of cleanup on this table so that can grow and become slowly over time

packages/shared/src/db/functions/learning-scenario.ts
```diff
@@ -422,6 +422,90 @@ export function dbGetSharedLearningScenarioConversations({
     .where(eq(activeShare.learningScenarioId, learningScenarioId));
 }
```
rlaerm, line 423: same points as for character:
- return value
- db index

packages/shared/src/learning-scenarios/learning-scenario-service.ts
```diff
@@ -421,6 +428,95 @@ export async function unshareLearningScenario({
...
+  }
+
+  const updatedShare = await dbExtendSharedLearningScenarioExpiration({
+    learningScenarioId,
+    user,
+    additionalTimeInMinutes,
+  });
+
+  if (!updatedShare) {
+    throw new NotFoundError('No active sharing found for this learning scenario');
```
rlaerm: same as in character-service: InvalidArgumentError

rlaerm (approved):

apps/chat-bot/src/components/custom-chat/custom-chat-share-with-learners/custom-chat-adjust-token-limit-button.tsx
brueckb: [packages/ui/src/components/select.tsx](https://github.com/FWU-DE/ais-chat/blob/main/packages/ui/src/components/select.tsx)  only accepts strings. I have made this a little more stringent by moving the conversion into the component, but it has to happen at some point :)

packages/shared/src/db/functions/character.ts
brueckb: added it to the cleanup ticket to not intervene with other migrations :)

PR #1092, opened 2026-07-01
Xberg

recursive-hub:
- added local xberg docker config
- added new file extensions
- removed old file extractors
- more explicit file handling in handleFileUpload


devops/docker/docker-compose.local.yml
TobiasWallura-xitaso, line 1: The container must be added to `docker-compose.yml` as well.

devops/docker/docker-compose.local.yml
```diff
@@ -96,6 +96,22 @@ services:
...
+    image: ghcr.io/xberg-io/kreuzberg:4.9.9-core
+    environment:
+      KREUZBERG_MAX_UPLOAD_SIZE_MB: '50'
+    ports:
+      - '8000:8000'
+    volumes:
+      - xberg_cache:/app/.kreuzberg
+    restart: unless-stopped
+    healthcheck:
+      test: ['CMD', 'kreuzberg', '--version']
```
TobiasWallura-xitaso: Health checks usually poll an endpoint. Maybe change to `curl http://localhost:8000/health` (that's how it's configured in k8s with the helm chart).

devops/docker/docker-compose.local.yml
```diff
@@ -96,6 +96,22 @@ services:
...
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 40s
```
TobiasWallura-xitaso: Should not be needed, docs say that xberg startup is fast (we used the delay for crawl4ai, which is quite slow).

apps/chat-bot/src/utils/files/generic.ts
```diff
@@ -1,44 +1,28 @@
...
-  const lastPart = parts[parts.length - 1];
-  if (lastPart === undefined) {
-    return fileName;
-  }
-
-  if (
-    !SUPPORTED_DOCUMENTS_EXTENSIONS.includes(lastPart.toString()) &&
-    !SUPPORTED_IMAGE_EXTENSIONS.includes(lastPart as SUPPORTED_IMAGE_TYPE)
-  ) {
+export function getFileExtension(fileName: string): string {
```
TobiasWallura-xitaso, line 7: This fn and `validateFileExtension` are nearly same, except for the list of file extensions and one throws the other returns a boolean. should be commonized.

apps/chat-bot/src/app/api/file-operations/preprocess-image.ts
```diff
@@ -58,7 +58,7 @@ export async function createImageAttachmentsForConversation(
 
 export async function preprocessImage(
   fileContent: Buffer,
-  type: SUPPORTED_DOCUMENTS_TYPE,
+  type: string,
```
TobiasWallura-xitaso, line 61: Why not keep the string literal type for improved type safety?

apps/chat-bot/src/app/api/file-extraction/file-extraction-xberg.ts
```diff
@@ -0,0 +1,53 @@
...
+    const response = await fetch(`${env.xbergUrl}/extract`, {
+      method: 'POST',
+      body: formData,
+      signal: AbortSignal.timeout(timeout),
+    });
+
+    if (!response.ok) {
+      const responseBody = await response.text();
+      throw new Error(
+        `Xberg request failed with status ${response.status} for file: ${filename}: ${responseBody.substring(0, 500)}`,
```
TobiasWallura-xitaso: For improved error monitoring, use sentry logging and pass in response details as additional data.

recursive-hub: When I log with sentry, then throw an exception, we would have two Sentry events, since the exception will be logged again. For now I would remove logging the responseBody due to the security concern from the other comment.

apps/chat-bot/src/app/api/file-operations/preprocess-image.ts
recursive-hub: I looked into this during implementation. The previous type SUPPORTED_DOCUMENTS_TYPE was wrong. Now the supported types are string arrays and the code is much simpler, e.g. SUPPORTED_FILE_EXTENSIONS.includes(lastPart) is now possible. This function only has one caller and isImageFile is already checked before. Making it completely typesafe would require a bigger change, e.g. "as const" on the arrays, a derived type, type guards for some more functions.

apps/chat-bot/src/app/api/file-extraction/file-extraction-xberg.ts
```diff
@@ -0,0 +1,53 @@
...
+      );
+    }
```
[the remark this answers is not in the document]
recursive-hub: done

devops/docker/docker-compose.local.yml
recursive-hub: done

devops/docker/docker-compose.local.yml
recursive-hub: Done

apps/chat-bot/src/utils/files/generic.ts
recursive-hub: I tried to commonize. Any unification either changes behaviour and need adjustments in other parts of the code (due to the try catch removal) or is cosmetic (shared 1-line helper). Would leave as is for now.

devops/docker/docker-compose.local.yml
recursive-hub: Done

apps/chat-bot/src/app/api/v1/files/route.ts
```diff
@@ -60,36 +61,44 @@ async function handleFileUpload(file: File) {
...
-  await dbInsertFileWithChunks(fileModel, chunks);
-  logDebug(`File ${file.name} with type ${fileExtension} stored in db.`);
+    await uploadMessageAttachment({ fileId, fileExtension, buffer: imageBuffer });
+    await dbInsertFileWithChunks(
+      {
+        id: fileId,
+        name: file.name,
+        size: imageBuffer.length,
+        type: fileExtension,
+        metadata,
```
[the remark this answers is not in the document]
recursive-hub: Done

TobiasWallura-xitaso (approved):

apps/chat-bot/src/app/api/file-extraction/file-extraction-xberg.ts
```diff
@@ -0,0 +1,50 @@
...
+    const markdown = results?.[0]?.content?.trim() ?? '';
+
+    if (!markdown) {
+      throw new Error(`Xberg returned no content for file: ${filename}`);
+    }
+
+    return markdown;
+  } catch (error) {
+    if (error instanceof Error && (error.name === 'AbortError' || error.name === 'TimeoutError')) {
+      throw new Error(`Xberg request timed out for file: ${filename}`);
```
rlaerm, line 45: It is very difficult to react on such errors individually. We should have at least one FileExtractionError so that we know in the frontend that there was a problem with the file processing. We can then show an individual message to the user. The error should contain the file name as separate value so that we can use it in the error message and the user knows which file was the problem.

apps/chat-bot/src/app/api/v1/files/route.ts
```diff
@@ -60,36 +61,52 @@ async function handleFileUpload(file: File) {
...
-    name: file.name,
-    size: extractResult.processedBuffer ? extractResult.processedBuffer.length : file.size,
-    type: fileExtension,
-    metadata: extractResult.metadata,
-    userId: user.id,
-  };
-  await dbInsertFileWithChunks(fileModel, chunks);
-  logDebug(`File ${file.name} with type ${fileExtension} stored in db.`);
+    await uploadMessageAttachment({ fileId, fileExtension: processedType, buffer: imageBuffer });
+    await dbInsertFileWithChunks(
```
rlaerm, line 78: I think we agreed on putting the business logic into service instances and not directly into route components and some utility functions. There should be a service instance in our shared package, maybe some abstractions, types, errors, etc.

PR #1095, opened 2026-07-06
TD-1334: add confirmation dialog to community sharing

stschaller-xitaso:


packages/ui/src/components/checkbox.tsx
```diff
@@ -6,10 +6,26 @@ import { Checkbox as CheckboxPrimitive } from 'radix-ui';
...
-function Checkbox({ className, ...props }: React.ComponentProps<typeof CheckboxPrimitive.Root>) {
+function Checkbox({
+  className,
+  onKeyDown,
+  ...props
+}: React.ComponentProps<typeof CheckboxPrimitive.Root>) {
+  const handleKeyDown: React.KeyboardEventHandler<HTMLButtonElement> = (event) => {
+    onKeyDown?.(event);
+
+    if (event.defaultPrevented || event.key !== 'Enter') {
```
TobiasWallura-xitaso, line 17: What about space key?

packages/ui/src/components/checkbox.tsx
```diff
@@ -6,10 +6,26 @@ import { Checkbox as CheckboxPrimitive } from 'radix-ui';
...
 import { CheckIcon } from 'lucide-react';
 
-function Checkbox({ className, ...props }: React.ComponentProps<typeof CheckboxPrimitive.Root>) {
+function Checkbox({
+  className,
+  onKeyDown,
+  ...props
+}: React.ComponentProps<typeof CheckboxPrimitive.Root>) {
+  const handleKeyDown: React.KeyboardEventHandler<HTMLButtonElement> = (event) => {
+    onKeyDown?.(event);
```
TobiasWallura-xitaso, line 15: Should this be below the `if`?

apps/chat-bot/src/components/custom-chat/custom-chat-share-section.tsx
```diff
@@ -31,6 +35,11 @@ export default function CustomShareSection<T extends FieldValues>({
 }: CustomShareSectionProps<T>) {
   const t = useTranslations('sharing');
   const toast = useToast();
+  const [isCommunityConfirmOpen, setIsCommunityConfirmOpen] = useState(false);
+  const { field: communityField } = useController({
+    name: communitySharingName,
+    control,
+  });
```
TobiasWallura-xitaso, line 42: fyi, at other places we use the `Controller` component. with that component the `Boolean` conversions (`Boolean(communityField.value)`) might not be needed.

apps/chat-bot/src/components/custom-chat/custom-share-confirmation-dialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+  onAccept: () => void;
+};
+
+export function CustomShareConfirmationDialog({
+  open,
+  onOpenChange,
+  onAccept,
+}: CustomShareConfirmationDialogProps) {
+  const t = useTranslations('sharing');
+  const checklistItemKeys = useMemo(
```
TobiasWallura-xitaso: no need for memo, could be a const on file level.

apps/chat-bot/src/components/custom-chat/custom-share-confirmation-dialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+        'community-confirmation.items.copyright-image-reviewed',
+        'community-confirmation.items.copyright-sources-provided',
+        'community-confirmation.items.personal-no-data',
+        'community-confirmation.items.personal-no-invitation',
+        'community-confirmation.items.ethics-no-real-person',
+        'community-confirmation.items.ethics-no-discrimination',
+        'community-confirmation.items.content-no-extremism-violence-sexualized',
+        'community-confirmation.items.content-no-false-outdated',
+        'community-confirmation.items.terms-accepted',
+      ] as const,
```
TobiasWallura-xitaso: Should we do a map for each section with it's entries? then some duplications below might be avoided (the ones with `checklistItemKeys.slice(x, y)`).
e.g.:

```suggestion
      {
        'community-confirmation.sections.copyright': [
          'community-confirmation.items.copyright-material-reviewed',
          'community-confirmation.items.copyright-no-protected-upload',
          'community-confirmation.items.copyright-image-reviewed',
          'community-confirmation.items.copyright-sources-provided',
        ],
        // ....
      } as const,
```

apps/chat-bot/messages/de.json
```diff
@@ -61,7 +61,36 @@
...
+        "personal-no-invitation": "Die Vorlage lädt nicht zur Eingabe personenbezogener oder ‑beziehbarer Daten ein.",
+        "ethics-no-real-person": "Die Vorlage ist nicht nach historischen oder realen Personen benannt.",
+        "ethics-no-discrimination": "Die Vorlage ist in ihrer Konzeption nicht diskriminierend.",
+        "content-no-extremism-violence-sexualized": "Die Vorlage enthält keine extremistischen, gewaltverherrlichenden oder sexualisierten Inhalte.",
+        "content-no-false-outdated": "Die Vorlage enthält keine falschen oder veralteten Inhalte (z.B. telli statt AIS.chat).",
+        "terms-accepted": "Ich habe die Nutzungsbedingungen gelesen und akzeptiere sie."
+      },
+      "terms-prefix": "Ich habe die",
+      "terms-link-label": "Nutzungsbedingungen",
+      "terms-suffix": "gelesen und akzeptiere sie."
```
TobiasWallura-xitaso: Can we merge these into a single localization key? e.g. use markdown to render the link similar to how it's done with the `faq-link`.
```suggestion
      "terms": "Ich habe die [Nutzungsbedingungen](https://ais-chat.schule/terms-of-use/) gelesen und akzeptiere sie."
```

TobiasWallura-xitaso: also: that would be the same as the `terms-accepted` key, one of them might not be needed.

TobiasWallura-xitaso (approved):

apps/chat-bot/src/components/custom-chat/custom-share-confirmation-dialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+        if (!nextOpen) {
+          resetChecklist();
+        }
+
+        onOpenChange(nextOpen);
+      }}
+    >
+      <AlertDialogContent
+        className="grid max-h-[calc(100dvh-1rem)] w-[calc(100%-1rem)] max-w-3xl grid-rows-[auto_minmax(0,1fr)_auto] gap-4 overflow-hidden p-4 pt-4 sm:max-h-[calc(100dvh-2rem)] sm:w-full sm:gap-6 sm:p-8 sm:pt-6"
```
rlaerm: Suggestion: All those class names are layout logic that could be extracted into a layout component. Maybe some of them can also go into the main comonent under shared/components. For all the other names there is a reason for that. Give it a name and provide a layout component that can be reused.

apps/chat-bot/src/components/custom-chat/custom-share-confirmation-dialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+          <AlertDialogTitle>{t('community-confirmation.title')}</AlertDialogTitle>
+        </AlertDialogHeader>
+
+        <div className="min-h-0 overflow-y-auto pr-1 sm:pr-2">
+          <div className="space-y-6">
+            <AlertDialogDescription>
+              {t('community-confirmation.description')}
+            </AlertDialogDescription>
+            <section className="space-y-3">
+              <h3 className="font-semibold">{t('community-confirmation.sections.copyright')}</h3>
```
rlaerm: h3 is only allowed if there is h1 and h2 because of accessibility. I don't know the reason why you choose h3 but I would suggest to use a normal div and add formatting if necessary.

rlaerm: There are multiple occurrences of h3 that must be fixed.

apps/chat-bot/src/components/custom-chat/custom-share-confirmation-dialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+              <div className="space-y-2 sm:space-y-3">
+                {checklistItemKeys.slice(0, 4).map((itemKey, index) => (
```
rlaerm: As already suggested, I would not use slice here but give all slices a name. If we add another point, all the indexes will change.

packages/ui/src/components/checkbox.tsx
stschaller-xitaso: Is already handled by Radix primitive component. Should we still leave Enter here or do we only want Space in that case?

packages/ui/src/components/checkbox.tsx
stschaller-xitaso: I took another look at it and having this line above the "if" enables the parent component that contains the checkbox to have its own logic by calling "event.defaultPrevented". We can remove that if we say the parent component should never have its own custom logic for the checkbox behavior and only react to Enter key (and Space key implemented by radix primitive)

apps/chat-bot/src/components/custom-chat/custom-share-confirmation-dialog.tsx
stschaller-xitaso: I can move some of it in the general AlertDialogContent and create a variant "scrollable" or such for the rest. I would prefer doing this in a separate PR because I want to check all used AlertDialogs so far and whether or not some classes for responsiveness etc, can also be moved to base class or removed from parent component alltogether

PR #1097, opened 2026-07-06
add grace window for sharing extension

brueckb:


apps/chat-bot/src/app/(authed)/(chat-bot)/characters/[characterId]/character-view.tsx
```diff
@@ -141,8 +139,12 @@ export function CharacterView({
           }
           return { success: false };
         }}
+        onPollShareData={async () =>
```
Comment (rlaerm) [inline apps/chat-bot/src/app/(authed)/(chat-bot)/characters/[characterId]/character-view.tsx]: You can call getCharacterShareDataAction() with await and async but you don't have to. It just adds an extra layer that does not change the behavior. I am fine with both.

packages/shared/src/characters/character-service.ts
```diff
@@ -392,6 +393,74 @@ export const deleteCharacter = async ({
...
+  user,
+}: {
+  characterId: string;
+  user: Pick<UserModel, 'id'>;
+}): Promise<void> => {
+  const share = await dbGetLatestManageableCharacterShare({ characterId, user });
+
+  if (share) {
+    throw new InvalidArgumentError(
+      'A share session already exists or has recently expired. Please extend the existing session instead of starting a new one.',
```
TobiasWallura-xitaso, line 459: Is the error in case of recently expired share really intended?
I could imagine, that starting a new session within 2 hours after the last session is a perfectly valid use case. e.g. sharing the same character in two different classes. In that case the teacher probably wants a new invite code for the second class, so the first class cannot continue chatting.

packages/shared/src/db/functions/character.ts
```diff
@@ -532,22 +534,94 @@ export async function dbLiftSuspensionOnCharacter({ characterId }: { characterId
...
+      expiredAt,
+    })
+    .returning();
+
+  return newShare;
+}
+
+/**
+ * Marks a character share as manually stopped.
+ * Returns `undefined` if the share does not exist.
```
rlaerm: I think returning null is the correct type because we looked at the db and there is no record to update. I think we already discussed about the difference between null and undefined. It is fine to me because the code base is already inconsistent but I would do it.

apps/chat-bot/src/app/(authed)/(chat-bot)/characters/[characterId]/character-view.tsx
```diff
@@ -108,18 +108,16 @@ export function CharacterView({
         budgetUsedBySharedChat={budgetUsedBySharedChat}
         maxBudget={maxBudget}
         onShare={async (data) => {
-          const result = await shareCharacterAction({
+          return await shareCharacterAction({
```
Comment (TobiasWallura-xitaso) [inline apps/chat-bot/src/app/(authed)/(chat-bot)/characters/[characterId]/character-view.tsx]: minor: no need for lambda body and `return` statement

apps/chat-bot/src/hooks/use-share-data-polling.ts
```diff
@@ -33,6 +39,8 @@ export function useShareDataPolling({
       }
     };
 
+    void poll();
```
rlaerm: I think this adds another call to fetch these values. IMHO the data is already fetched with the character or learning scenario and we directly fetch it again.

apps/chat-bot/src/components/custom-chat/custom-chat-share-with-learners/custom-chat-grace-window-note.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }, 1000);
+    return () => clearInterval(timer);
+  }, [shareSessionState]);
+
+  if (shareSessionState !== ShareSessionState.EXPIRED_RECENTLY) return null;
+
+  return (
+    <p className="text-sm text-gray-600 text-center">
+      {t('session-will-be-closed', {
+        time: new Date(gracePeriodTimeLeftInSeconds * 1000).toISOString().slice(11, 19),
```
TobiasWallura-xitaso: What about the usage time of multiple days? that should still be rendered in a sensible way.

apps/chat-bot/e2e/tests/character/share-with-learners-grace-window.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      await expect(page.getByTestId('stop-share-button')).toBeVisible();
+    });
+
+    test('can manually stop a share session', async ({ page }) => {
+      await page.goto(`/characters/editor/${characterId}`);
+      await page.waitForURL(`**/characters/editor/${characterId}**`);
+
+      const stopShareButton = page.getByTestId('stop-share-button');
+
+      if (await stopShareButton.isVisible()) {
```
TobiasWallura-xitaso, line 55: With this `if` condition the test might incorrectly pass (the test must click the stop button, right?)

TobiasWallura-xitaso (approved):

rlaerm (approved):

apps/chat-bot/src/app/(authed)/(chat-bot)/characters/[characterId]/character-view.tsx
Comment (brueckb) [reply apps/chat-bot/src/app/(authed)/(chat-bot)/characters/[characterId]/character-view.tsx]: removed :)

apps/chat-bot/src/components/custom-chat/custom-chat-share-with-learners/custom-chat-grace-window-note.tsx
brueckb: this is about the grace window which is hardcoded to 2h. But i get it, will be addressed :)

TobiasWallura-xitaso: ah, understood. then it's fine as is.

packages/shared/src/characters/character-service.ts
brueckb: we explicitly decided against having two sharing sessions that overlap. We also do not have UI features that support handling several sessions. ;)

packages/shared/src/db/functions/character.ts
brueckb: yes and I agree. this is an AI based reintroduction, sorry for that! Will be fixed :)

PR #1098, opened 2026-07-06
Feat/td 1370 show dialog

rlaerm:


rlaerm:
commits were cherry picked into branch feat/TD-1370-show-dialog-instead-of-error

PR #1103, opened 2026-07-07
Remove old Azure endpoint functions and switch everything to the Responses API

AsamMax:


packages/ai-core/src/chat/providers/azure.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    },
+  } as AiModel;
+}
+
+describe('Azure chat providers', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('passes temperature and model additional parameters to Responses stream and maps usage', async () => {
```
TobiasWallura-xitaso: we no longer pass temperature, right?

packages/ai-core/src/chat/providers/azure.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      },
+    );
+    expect(onComplete).toHaveBeenCalledWith({
+      promptTokens: 10,
+      completionTokens: 20,
+      totalTokens: 30,
+    });
+  });
+
+  it('passes model additional parameters to Responses generation', async () => {
```
TobiasWallura-xitaso: What's the difference to the other test? should they be merged (at least from the name, they look very similar).

TobiasWallura-xitaso (approved):

PR #1110, opened 2026-07-08
TD-1394: fix hydration warnings

stschaller-xitaso:


apps/chat-bot/src/app/layout.tsx
```diff
@@ -56,13 +55,10 @@ export default async function RootLayout({ children }: { children: React.ReactNo
...
-        <Script
-          id="public-config"
-          // runs as soon as the browser parses it (before client components hydrate)
-          strategy="beforeInteractive"
-          dangerouslySetInnerHTML={{ __html: inlineScript }}
-        />
+      <head>
+        <script id="public-config" dangerouslySetInnerHTML={{ __html: inlineScript }} />
+      </head>
+      <body suppressHydrationWarning>
```
stschaller-xitaso, line 59: 1. React 19 changed how it handles script, script tags rendered inside the component body are no longer executed on the client -> plain script tags are handled correctly and runs before any hydration (removing strategy "beforeInteractive" is the same behavor for a simple inline config script like we use here)

2. <suppressHydrationWarning> in the body will suppress warnings which are caused by browser extensions, since it is only suppressing in the body, real hydration errors in the children will still be shown

apps/admin/src/app/layout.tsx
```diff
@@ -17,12 +16,7 @@ export default async function RootLayout({ children }: { children: ReactNode })
   return (
     <html lang="de" className={barlow.className}>
       <body style={{ margin: 0, padding: 0, fontFamily: 'sans-serif' }}>
-        <Script
-          id="public-config"
-          // runs as soon as the browser parses it (before client components hydrate)
-          strategy="beforeInteractive"
-          dangerouslySetInnerHTML={{ __html: inlineScript }}
-        />
+        <script id="public-config" dangerouslySetInnerHTML={{ __html: inlineScript }} />
```
stschaller-xitaso: React 19 changed how it handles script, script tags rendered inside the component body are no longer executed on the client -> plain script tags are handled correctly and runs before any hydration (removing strategy "beforeInteractive" is the same behavor for a simple inline config script like we use here)

apps/chat-bot/src/components/hooks/use-breakpoints.ts
```diff
@@ -15,7 +15,7 @@ const getWindowWidth = () => {
 };
 
 export default function useBreakpoints() {
-  const [width, setWidth] = useState(getWindowWidth());
+  const [width, setWidth] = useState(0);
```
stschaller-xitaso, line 19: On server getWindowWidth() returns 0 (window does not exist), on the client (hydration render) that window does exist and returns a value bigger than 0 causing a mismatch. Now both client and server start at 0 and after hydration useEffect rerenders the component with the correct viewport

apps/chat-bot/src/app/layout.tsx
```diff
@@ -56,13 +55,10 @@ export default async function RootLayout({ children }: { children: React.ReactNo
...
-      <body>
-        <Script
-          id="public-config"
-          // runs as soon as the browser parses it (before client components hydrate)
-          strategy="beforeInteractive"
-          dangerouslySetInnerHTML={{ __html: inlineScript }}
-        />
+      <head>
+        <script id="public-config" dangerouslySetInnerHTML={{ __html: inlineScript }} />
+      </head>
```
TobiasWallura-xitaso: What about this?
https://nextjs.org/docs/messages/no-script-tags-in-head-component

apps/chat-bot/src/components/hooks/use-breakpoints.ts
TobiasWallura-xitaso: maybe add a comment for future reference, e.g.
```suggestion
  // width is initialized with 0 instead of the current window width of the client to prevent hydration errors on the client
  const [width, setWidth] = useState(0);
```

TobiasWallura-xitaso (approved):

rlaerm (approved):

PR #1123, opened 2026-07-09
Feat/td 1378 clean up budget extension

stschaller-xitaso:


apps/chat-bot/src/components/custom-chat/share-with-learners/custom-chat-extend-share-expiration-dialog.tsx
```diff
@@ -70,7 +78,7 @@ export function CustomChatExtendShareExpirationDialog({
 
   return (
     <Dialog open={open} onOpenChange={handleOpenChange}>
-      <DialogTrigger asChild>{trigger}</DialogTrigger>
+      <DialogTrigger asChild>{triggerWithState}</DialogTrigger>
```
TobiasWallura-xitaso: Can we set `disabled` prop directly on `DialogTrigger` instead of cloning the trigger react element?

apps/chat-bot/src/components/custom-chat/share-with-learners/custom-chat-share-with-learners.tsx
```diff
@@ -136,6 +138,14 @@ export function CustomChatShareWithLearners({
   const currentBudgetUsedBySharedChat =
     polledData?.budgetUsedBySharedChat ?? budgetUsedBySharedChat;
 
+  const tokenLimitInCent =
+    maxBudget === null ? null : (maxBudget * currentTokenPointsPercentageLimit) / 100;
+  const currentTokenLimitExceeded =
+    polledData?.tokenLimitExceeded ??
+    (tokenLimitInCent !== null && currentBudgetUsedBySharedChat >= tokenLimitInCent);
```
TobiasWallura-xitaso: Can we remove the fallback?
```suggestion
```

packages/shared/src/sharing/const.ts
```diff
@@ -18,3 +18,7 @@ const SHARE_EXTENSION_WINDOW_HOURS = 2;
  * Value: 2 hours = 7,200,000 milliseconds
  */
 export const SHARE_EXTENSION_WINDOW_MS = SHARE_EXTENSION_WINDOW_HOURS * 60 * 60 * 1000;
+
+export const MAX_SHARE_USAGE_TIME_LIMIT_IN_MINUTES = 130 * 24 * 60;
+
+export const MAX_EXTENDABLE_USAGE_TIME_IN_SECONDS = 100 * 24 * 60 * 60; // 100 days
```
TobiasWallura-xitaso, line 24: This file contains everything, milliseconds, seconds, minutes and hours 😆

TobiasWallura-xitaso (approved):

packages/shared/src/sharing/const.ts
stschaller-xitaso: It contains multitudes 😂 But there is a reason for everything

PR #1152, opened 2026-07-16
Feat/td 975 refactor session storage data

rlaerm:


apps/chat-bot/src/components/shared-chat/learning-scenario-shared-chat.tsx
```diff
@@ -13,10 +14,17 @@ export default function LearningScenarioSharedChat({
   const t = useTranslations('learning-scenarios.shared');
   const { id, inviteCode, modelId } = sharedSchoolChat;
 
+  const initialMessages =
+    loadSharedChat(inviteCode)?.messages.map((message) => ({
+      id: message.id,
+      role: message.role,
+      content: message.content,
+    })) ?? [];
```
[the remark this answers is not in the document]
rlaerm: not necessary because react compiler is doing it

apps/chat-bot/src/components/shared-chat/learning-scenario-shared-chat.tsx
```diff
@@ -13,10 +14,17 @@ export default function LearningScenarioSharedChat({
...
+
```
stschaller-xitaso, line 23: If the array from loadSharedChat is simply empty this would "overwrite" the intialMessage fallback. In case of emtpy message array we would still like to show the initialMessage, right?

stschaller-xitaso (approved):

apps/chat-bot/src/components/shared-chat/learning-scenario-shared-chat.tsx
rlaerm: There is no initial message for learning scenarios. Learning scenarios do have an "Arbeitsauftrag" but no initial message. Characters have an initial message but no "Arbeitsauftrag".

PR #1166, opened 2026-07-20
show warning if page is expired

rlaerm:


packages/shared/src/db/schema.ts
```diff
@@ -943,6 +944,20 @@ export type LearningScenarioOptionalShareDataModel = z.infer<
   typeof learningScenarioOptionalShareDataModel
 >;
 
+/**
+ *  Type guard for learning scenario with share data
+ *  This uses Zod validation for compile-time and runtime safety.
```
TobiasWallura-xitaso, line 949: From what I see, this function is only used for validating an entity, which is loaded from database; in that case we don't need zod validation, at least IMO.

packages/shared/src/db/functions/learning-scenario.ts
```diff
@@ -275,18 +301,21 @@ export async function dbGetLearningScenariosByIds({
...
   learningScenarioId,
   user,
 }: {
   learningScenarioId: string;
   user: Pick<UserModel, 'id'>;
-}): Promise<LearningScenarioWithShareDataModel | undefined> {
-  const activeShare = latestActiveLearningScenarioShare(user);
-  const [row] = await baseLearningScenarioWithShareQuery(activeShare)
-    .innerJoin(activeShare, eq(activeShare.learningScenarioId, learningScenarioTable.id))
+}): Promise<LearningScenarioOptionalShareDataModel | undefined> {
```
TobiasWallura-xitaso: The intention of this function was to return a type, which always includes share data. If you need the learning scenario with optional share data, use `dbGetLearningScenarioByIdOptionalShareData` instead.

packages/shared/src/db/schema.ts
rlaerm: The function is needed for type safety. Otherwise Typescript complains about the type.

TobiasWallura-xitaso: Yes, because the db function now returns `LearningScenarioOptionalShareDataModel` instead of `LearningScenarioWithShareDataModel`. IMO that is the issue.

packages/shared/src/db/functions/learning-scenario.ts
rlaerm: I see. This function is only used by getSharedLearningScenario() which is my business logic that I want to change. The other function is used by the edit view which I do not want to change and does not return the same. But maybe it is just the wrong type. When I change the left join to an inner join it may work.

packages/shared/src/db/schema.ts
rlaerm: Okay, you are right. The guard is indeed not mandatory.

brueckb (approved):

PR #1167, opened 2026-07-20
TD-293: Include chat images in Word export

sammo3332:


apps/chat-bot/src/app/api/download-conversation/utils.ts
```diff
@@ -6,30 +6,41 @@ import {
...
 import { logError } from '@shared/logging';
 import { dbGetModelByName } from '@shared/db/functions/llm-model';
+import { getFileFromS3 } from '@shared/s3';
+import { getFileExtension, isImageFile } from '@/utils/files/generic';
+import { Readable } from 'stream';
+import sharp from 'sharp';
 
 const USER_FULL_NAME = 'Nutzer/in';
+const EXPORTED_IMAGE_MAX_HEIGHT = 256;
+type ExportedImageType = 'jpg' | 'png' | 'gif' | 'bmp';
```
TobiasWallura-xitaso: We only support these image formats, no need for gif and bmp.
```
export const SUPPORTED_IMAGE_EXTENSIONS = ['jpeg', 'jpg', 'png', 'svg', 'webp'] as const;
```

packages/shared/src/conversation/conversation-service.ts
```diff
@@ -147,14 +150,29 @@ export async function getConversationMessageForExport({
...
+  const messages = await getConversationMessages({ conversationId, userId });
+  const messageIndex = messages.findIndex(
+    (conversationMessage) => conversationMessage.id === messageId,
+  );
+  const previousUserMessage =
+    message.role === 'assistant'
+      ? messages
+          .slice(0, messageIndex)
+          .toReversed()
+          .find((conversationMessage) => conversationMessage.role === 'user')
```
TobiasWallura-xitaso: fyi, instead of `toReversed()` and `find()`, you can use `findLast()`.

packages/shared/src/conversation/conversation-service.ts
```diff
@@ -147,14 +150,29 @@ export async function getConversationMessageForExport({
...
 
   if (!message) {
     throw new NotFoundError('Conversation message not found');
   }
 
+  const messages = await getConversationMessages({ conversationId, userId });
+  const messageIndex = messages.findIndex(
+    (conversationMessage) => conversationMessage.id === messageId,
+  );
+  const previousUserMessage =
```
TobiasWallura-xitaso: Why do we need the "previousUserMessage"? When exporting a single assistant message, I assume that the uploaded images should not be included (same as with the user prompt).

apps/chat-bot/src/app/api/download-conversation/route.ts
TobiasWallura-xitaso, line 1: With the ticket TD-975, it is now possible to upload files in shared chats as well -> downloading a shared chat conversation should probably also include the uploaded images.

packages/shared/src/conversation/conversation-service.ts
sammo3332: This code  was removed after the other review comment.

apps/chat-bot/src/app/api/download-conversation/utils.ts
sammo3332: God point, it is adjusted .

sammo3332:
@copilot resolve the merge conflicts in this pull request

apps/chat-bot/src/components/chat/shared-header-bar.tsx
```diff
@@ -122,11 +132,12 @@ export function SharedChatHeader({
       {!showCompactHeader ? (
         <>
           <DownloadSharedConversationButton
-            conversationMessages={messages}
+            conversationMessages={messagesWithFiles}
```
recursive-hub, line 135: Could we avoid deriving messagesWithFiles in shared-header-bar? This leaks export-specific file metadata into the header UI. It is cleaner to keep messages as-is here and have the download flow resolve the file mapping when the user clicks export.

apps/chat-bot/src/components/common/expired-chat-modal.tsx
```diff
@@ -17,16 +18,18 @@ import {
 import { ArrowClockwiseIcon } from '@phosphor-icons/react';
 
 type ExpiredChatModalProps = {
-  conversationMessages: Message[];
+  conversationMessages: SharedConversationMessage[];
```
recursive-hub, line 21: Introducing SharedConversationMessage into ExpiredChatModal seems to push export-specific data into a core UI component. Can you resolve the files closer to the download boundary?

apps/chat-bot/src/app/api/download-conversation/utils.ts
```diff
@@ -89,33 +100,40 @@ function getConversationMetadata({ conversation }: { conversation: ConversationM
...
+            bold: true,
+            size: 22,
+          }),
+        ],
+      }),
+      ...markdownToDocx(message.content),
+      ...(await getImageParagraphsForMessage({ messageId: message.id, fileMapping })),
+      new Paragraph({}),
+    ]),
+  ).then((messageSections) => messageSections.flat());
```
TobiasWallura-xitaso, line 127: minor, can we use `await` here instead of chaining `.then(...)`? The function is already `async`, so `const messageSections = await Promise.all(...); return messageSections.flat();` reads more consistently.

apps/chat-bot/src/app/api/download-conversation/shared/utils.ts
```diff
@@ -63,23 +68,28 @@ type SectionType = Paragraph | Table;
...
+            bold: true,
+            size: 22,
+          }),
+        ],
+      }),
+      ...markdownToDocx(message.content),
+      ...(await getImageParagraphsForMessage({ messageId: message.id, fileMapping })),
+      new Paragraph({}),
+    ]),
+  ).then((messageSections) => messageSections.flat());
```
TobiasWallura-xitaso, line 92: minor, can we use `await` here instead of chaining `.then(...)`? The function is already `async`, so `const messageSections = await Promise.all(...); return messageSections.flat();` reads more consistently.

apps/chat-bot/src/app/api/download-conversation/utils.ts
```diff
@@ -166,6 +185,126 @@ export async function generateConversationMessageDocxFile({
...
+
+  const scale = Math.min(1, EXPORTED_IMAGE_MAX_HEIGHT / height);
+
+  return {
+    width: Math.round(width * scale),
+    height: Math.round(height * scale),
+  };
+}
+
+async function streamToBuffer(stream: Readable): Promise<Buffer> {
```
TobiasWallura-xitaso, line 298: We already have an exported `streamToBuffer` in `@/utils/files/image-data` (used by `scaled-image-service.ts` and `image-generation-input-files.ts`). Can we reuse it here instead of re-implementing it? The existing one also handles string chunks.

apps/chat-bot/src/app/api/download-conversation/shared/route.ts
```diff
@@ -52,6 +72,93 @@ export async function POST(request: NextRequest) {
...
+
+    if (messageFiles.length > 0) {
+      fileMapping.set(messageId, messageFiles);
+    }
+  }
+
+  return fileMapping;
+}
+
+function verifySharedChatFilesBelongToSession({
```
TobiasWallura-xitaso, line 132: Since invite codes are unique per entity, `inviteCode` + `sessionId` already pins a file to one entity, so the `entityType`/`entityId` checks here are redundant. Can we simplify `getSharedChatFileMapping` to verify only that each file is shared-chat metadata matching `inviteCode` + `sessionId` (plus the existing `verify.filesDoNotBelongToAnyUser`)? That lets us drop the `getSharedChatEntity` fetch, the entity derivation in `verifySharedChatFilesBelongToSession`, and the follow-up `verify.sharedChatFileOwnershipBySession` call — collapsing the two overlapping ownership checks into one.

apps/chat-bot/src/app/api/download-conversation/shared/route.ts
```diff
@@ -52,6 +72,93 @@ export async function POST(request: NextRequest) {
...
+  const files = await dbGetFilesInIds(fileIds);
+  verify.filesDoNotBelongToAnyUser(files);
+  const fileMetadata = verifySharedChatFilesBelongToSession({ files, inviteCode, sharedSessionId });
+  const sharedEntity = await getSharedChatEntity({
+    inviteCode,
+    entityType: fileMetadata.entityType,
+    entityId: fileMetadata.entityId,
+  });
+
+  verify.sharedChatEntityIsAccessible(sharedEntity);
```
TobiasWallura-xitaso, line 107: `verify.sharedChatEntityIsAccessible` throws on expired/deleted/suspended/stopped chats, but `ExpiredChatModal` exists specifically to allow downloading *after* expiry, and the text-only export path only calls `checkInviteCodeForExport` (no accessibility gate). As-is, an expired chat with images would fail to export while a text-only one succeeds. Is this gate intended? If not, it can be dropped along with the `sharedEntity` fetch (see the related comment on `getSharedChatFileMapping`).

TobiasWallura-xitaso (changes requested):

apps/chat-bot/src/app/api/download-conversation/shared/route.ts
sammo3332: Good point, I simplified the check and removed the duplicate validation.

apps/chat-bot/src/app/api/download-conversation/utils.ts
sammo3332: I reused the existing helper and removed the duplicate function.

apps/chat-bot/src/components/chat/shared-header-bar.tsx
sammo3332: Good point. I moved the file mapping into the download flow, so the header now works with the original messages.

apps/chat-bot/src/components/common/expired-chat-modal.tsx
sammo3332: Done. The modal now receives regular chat messages, and the files are resolved when the download starts.

TobiasWallura-xitaso (approved):

PR #1170, opened 2026-07-21
TD-1405: (For e2e) Route seeded models through Bifrost

AsamMax:


.github/workflows/e2e.yml
```diff
@@ -122,6 +122,27 @@ jobs:
...
+          elapsed=0
+          until wget -qO- http://localhost:8089/api/providers > /dev/null 2>&1; do
+            if [ $elapsed -ge $timeout ]; then
+              echo "Bifrost did not become ready within ${timeout}s"
+              exit 1
+            fi
+            echo "$(date) - waiting for Bifrost to be ready"
+            sleep 1
+            elapsed=$((elapsed + 1))
+          done
```
TobiasWallura-xitaso: Can we leave that out and let it start in the background?
If it really takes a long time to start, prefer to do the waiting at a later step (e.g. right before the seeding).

TobiasWallura-xitaso: in #1185 bifrost was added to e2e.yml as well, sorry for the conflict

.github/workflows/e2e.yml
```diff
@@ -122,6 +122,27 @@ jobs:
         run: |
           node devops/docker/mock-llm/server.mjs > /tmp/mock-llm.log 2>&1 &
 
+      - name: Start Bifrost container
+        run: |
+          docker run \
+            --name bifrost \
+            --add-host=host.docker.internal:host-gateway \
```
TobiasWallura-xitaso: as a simpler solution: use `--network host`, then bifrost can keep using "localhost" when connecting to the mock llm.

apps/chat-bot/e2e/utils/chat.ts
```diff
@@ -20,18 +20,24 @@ export async function sendMessage(page: Page, message: string) {
...
-    await waitForLoadingSpinner;
-    // Wait for the loading spinner to disappear, which indicates that the response has started streaming
-    await loadingSpinner.waitFor({ state: 'detached', timeout: 60_000 });
+
+    // Firefox can miss the short-lived loading animation for fast mock responses.
+    // Keep waiting for completion, but only wait for spinner detachment if it was observed.
+    const didShowLoadingSpinner = await loadingSpinner
+      .waitFor({ timeout: 5_000 })
+      .then(() => true)
+      .catch(() => false);
```
TobiasWallura-xitaso: The promise must be created before `press('Enter')`;
After `press('Enter')` it can be awaited.

apps/chat-bot/e2e/utils/chat.ts
```diff
@@ -20,18 +20,24 @@ export async function sendMessage(page: Page, message: string) {
...
+      await loadingSpinner.waitFor({ state: 'detached', timeout: 60_000 });
+    }
 
     // Either the response finishes successfully and shows the Reload button,
     // or an error message appears and the test should fail.
     await Promise.race([
-      reloadButton.waitFor({ timeout: 20_000 }),
-      errorText.waitFor({ timeout: 20_000 }).then(() => {
+      reloadButton.waitFor({ timeout: 60_000 }),
+      errorText.waitFor({ timeout: 60_000 }).then(() => {
```
TobiasWallura-xitaso: please revert

packages/api-database/src/seed.ts
```diff
@@ -279,8 +279,39 @@ export async function seedDatabase() {
...
+    const seededModels: LlmModel[] = [];
+    await db.delete(llmModelApiKeyMappingTable).where(
+      and(
+        eq(llmModelApiKeyMappingTable.apiKeyId, apiKey.id),
+        inArray(
+          llmModelApiKeyMappingTable.llmModelId,
+          DEFAULT_MODELS.map((model) => model.id!),
+        ),
+      ),
+    );
```
TobiasWallura-xitaso, line 291: Why do we delete the mappings first?

TobiasWallura-xitaso (approved):

.github/workflows/e2e.yml
AsamMax: Applied. Bifrost now starts once in the background, and the readiness wait runs immediately before API database seeding. The merge also retains the v1.6.5 image from #1185.

.github/workflows/e2e.yml
AsamMax: Applied. The CI Bifrost container now uses `--network host`, and `LLM_MOCK_BASE_URL` uses `http://localhost:6556`.

apps/chat-bot/e2e/utils/chat.ts
AsamMax: Reverted. `sendMessage` is back to the original spinner and 20-second completion waits.

apps/chat-bot/e2e/utils/chat.ts
AsamMax: Applied. The loading-spinner promise is created before pressing Enter and awaited afterwards.

packages/api-database/src/seed.ts
AsamMax: The mapping table has no unique constraint on `(llmModelId, apiKeyId)`, so `onConflictDoNothing()` does not prevent duplicate mappings when the seed is rerun. The delete is scoped to this seeded API key and the default seeded model IDs, then those mappings are recreated; unrelated mappings are preserved.

.github/workflows/e2e.yml
```diff
@@ -271,6 +285,12 @@ jobs:
           echo "=== mock-llm logs ==="
           cat /tmp/mock-llm.log || true
 
+      - name: Print Bifrost logs
+        if: always()
+        run: |
+          echo "=== Bifrost logs ==="
+          cat /tmp/bifrost.log || true
+
```
TobiasWallura-xitaso:
```suggestion
```

PR #1191, opened 2026-07-24
TD 412: refactor locale share chats

brueckb:


TobiasWallura-xitaso (approved):

rlaerm (approved):

rlaerm:
Fine for now, but we have to align on the wording. In packages/shared/src/entities/entity-types we have already those entityTypes and now we also have customChatVariants. That is something we should solve soon.

PR #1193, opened 2026-07-27
Feat/td 1441 create url presets table

rlaerm:


packages/shared/src/web-search/url-presets/types.ts
```diff
@@ -0,0 +1,17 @@
+import z from 'zod';
+
+export const urlSchema = z.httpUrl();
+
+export const urlPresetSchema = z.object({
+  id: z.uuidv4().nonoptional(),
```
TobiasWallura-xitaso: we use `uuid()` in other places, why the explicit uuidv4 here?

packages/shared/src/web-search/url-presets/url-preset-admin-service.ts
```diff
@@ -0,0 +1,45 @@
+import { eq } from 'drizzle-orm';
+import { db } from '@shared/db';
+import { urlPresetTable } from '@shared/db/schema';
+import { UrlPreset, UrlPresetInsert, UrlPresetUpdate } from './types';
+import { NotFoundError } from '@shared/error';
+
+export async function getAllUrlPresets(): Promise<UrlPreset[]> {
+  const presets = await db.select().from(urlPresetTable).orderBy(urlPresetTable.orderNumber);
```
TobiasWallura-xitaso: Should we add a fallback orderBy (e.g. by name), in case the orderNumber is the same?

TobiasWallura-xitaso (approved):

packages/shared/src/web-search/url-presets/types.ts
rlaerm: because postgresql uses uuid v4 and it is more specific. But I can also switch back to the common uuid function, thats fine for me.

packages/shared/src/web-search/url-presets/url-preset-admin-service.ts
rlaerm: yes, why not. I add the name as second criteria

packages/shared/src/db/schema.ts
```diff
@@ -385,6 +385,8 @@ export const webSearchScopeSchema = z.enum(['all-web', 'included-domains']);
 export const webSearchScopeEnum = pgEnum('web_search_scope', webSearchScopeSchema.enum);
 export type WebSearchScope = z.infer<typeof webSearchScopeSchema>;
 
+export const webSearchIncludedDomainsSchema = z.array(z.httpUrl());
```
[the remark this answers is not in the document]
rlaerm: z.httpUlr() is a known function of the zod library and does not need an augmentation.

PR #1194, opened 2026-07-27
TD-1442: Fix filterGroup type and cleanup

TobiasWallura-xitaso:
Fixes a bug where the DB could contain `{}` in the `filter_attributes` column, while the TypeScript type declared all properties as non-optional with a default of `[]`. This mismatch meant downstream code could crash on records with missing filter fields (e.g. older customChats).

This PR fixes the typings so that all `FilterGroup` properties are optional, matching what's actually persisted in the DB. As a result, the `createEmptyFilterGroup` fallback helper is no longer needed and has been removed.

Also simplifies filterGroup construction in the assistant/character/learning-scenario editors by passing form data directly to `toFilterGroup`, and drops unused filter helpers (`hasActiveFilters`, `getActiveFilterPillLabels`).


apps/chat-bot/src/i18n/sharing-locale.ts
```diff
@@ -71,7 +71,7 @@ export async function resolveSharingLocale({
       customChatId: character.id,
       ownerUserId: character.userId,
       locale,
-      existingFilterGroup: character.filterGroup ?? createEmptyFilterGroup(),
+      existingFilterGroup: character.filterGroup,
```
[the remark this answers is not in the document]
TobiasWallura-xitaso: This is fine, db typings ensure the `filterGroup` is non-null.

packages/shared/src/db/schema.ts
```diff
@@ -521,14 +521,14 @@ export const subjectsSchema = z.enum([
...
-  subjects: z.array(subjectsSchema).default([]),
-  categories: z.array(categoriesSchema).default([]),
-  federal_states: z.array(federalStatesSchema).default([]),
-  languages: z.array(languagesSchema).default([]),
+  school_types: z.array(schoolTypesSchema).optional(),
+  grade_ranges: z.array(gradeRangesSchema).optional(),
+  subjects: z.array(subjectsSchema).optional(),
+  categories: z.array(categoriesSchema).optional(),
+  federal_states: z.array(federalStatesSchema).optional(),
+  languages: z.array(languagesSchema).optional(),
```
AsamMax, line 529: I personally think empty arrays are more clear than null values. But that doesn't have to be part of this pr

AsamMax (approved):

packages/shared/src/db/schema.ts
TobiasWallura-xitaso: Yes, agree. However the DB will contain `{}` (i.e. empty object) for all entities created before the filters were introduced. Therefore we need to make sure to align our TS type to that case.
Transforming objects after reading from db might be possible as well, but that's usually not how we handle it.

PR #1202, opened 2026-07-28
TD-1445: Make static models configurable

AsamMax:
Add a shared chat-app configuration table and a dedicated admin view for assigning the global chat, fallback, auxiliary, and image model roles.

image

image


packages/shared/src/llm-models/llm-model-service.ts
```diff
@@ -1,13 +1,31 @@
...
  * Get the default model from a list of models
  * @param models The list of LLM models
  * @returns The default model or undefined if none found
  */
-export function getDefaultModel(models: LlmModelSelectModel[]): LlmModelSelectModel | undefined {
-  return models.find((m) => m.name === DEFAULT_CHAT_MODEL) ?? getFirstTextModel(models);
+export function getDefaultModel<T extends LlmModelWithOptionalStaticRoles>(
+  models: T[],
+): T | undefined {
+  return getModelWithRole(models, 'default-chat') ?? (getFirstTextModel(models) as T | undefined);
```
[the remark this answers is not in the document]
AsamMax: This is Intentionally out of scope. There will be a seperate PR with circuit-breaker logic

apps/admin/src/app/ais-chat-app/static-models/actions.ts
```diff
@@ -0,0 +1,46 @@
...
+    (configurations) =>
+      new Set(configurations.map((configuration) => configuration.role)).size ===
+      configurations.length,
+    'Each static model role must be configured exactly once.',
+  );
+
+export async function getStaticModelConfigurationAction() {
+  await requireAdminAuth();
+  return runServerAction('getStaticModelConfigurationAction', async () => {
+    const [models, configurations] = await Promise.all([
```
rlaerm: Until now we use actions only as a very thin wrapper for service calls and put the business logic/validation/db calls into those service classes. I would stick with that pattern.

apps/admin/src/app/ais-chat-app/static-models/StaticModelConfigurationView.tsx
```diff
@@ -0,0 +1,103 @@
...
+                  .map((model) => (
+                    <SelectItem key={model.id} value={model.id}>
+                      {model.displayName} ({model.name})
+                    </SelectItem>
+                  ))}
+              </SelectContent>
+            </Select>
+          </div>
+        ))}
+        <Button onClick={() => void save()} disabled={isSubmitting || !isComplete}>
```
rlaerm, line 104: There is a CardFooter that can be used and also a CardRow instead of the div above. Maybe it helps.

packages/shared/migrations/0086_material_deathstrike.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE "static_model_configuration" (
```
rlaerm: I know it is a little bit late but here is how I would have done it:
1. Discuss with the team about how to do it. Maybe there are some interesting ideas and suggestions.
2. The name is a little bit weired, static and configuration? I would name it default_models.
3. I would add another column for the federalStateId which is nullable. Not much work but then we can override the default models for a federal state. But we can do this afterwards, too.
4. The implementation is very specific. Why not generalize the problem. It is about a configuration that rarely changes. Instead of model_id we could use a json as we did with some other configurations as well. In this case we could reuse the table and the db calls and the service to store whatever configuration we want for the whole project as well as for federal states.

I realize that no one has the time or inclination to change this anymore. But maybe next time.

packages/shared/src/db/functions/character.ts
```diff
@@ -246,7 +245,7 @@ export async function dbCreateCharacter(
 ) {
   let modelId = character.modelId;
   if (!modelId) {
-    modelId = (await dbGetModelByName(DEFAULT_CHAT_MODEL))?.id;
+    modelId = (await dbGetStaticModelByRole('default-chat'))?.id;
     if (!modelId) {
       throw new Error('No default model found');
```
rlaerm, line 250: I would use a NotFoundError here

packages/shared/src/db/functions/llm-model.ts
```diff
@@ -53,6 +59,65 @@ export async function dbGetLlmModelsByFederalStateId({
...
+    )
+    .where(
+      and(
+        eq(staticModelConfigurationTable.role, role),
+        eq(federalStateLlmModelMappingTable.federalStateId, federalStateId),
+        eq(llmModelTable.isDeleted, false),
+      ),
+    )
+    .$withCache();
+  return model;
```
rlaerm, line 78: getById() functions where we expect a record to be returned should check if there is a record and throw a NotFoundError otherwise. I know the old codebase is not doing that, but new functions should follow this pattern.

packages/shared/src/db/functions/llm-model.ts
```diff
@@ -53,6 +59,65 @@ export async function dbGetLlmModelsByFederalStateId({
...
+    .select({ role: staticModelConfigurationTable.role, model: llmModelTable })
+    .from(staticModelConfigurationTable)
+    .innerJoin(llmModelTable, eq(staticModelConfigurationTable.modelId, llmModelTable.id))
+    .$withCache();
+}
+
+export async function dbUpdateStaticModelConfigurations(
+  configurations: { role: StaticModelRole; modelId: string }[],
+) {
+  return db.transaction(async (tx) => {
```
rlaerm: That speaks in favor of a JSON column that holds all default models.

packages/shared/src/image-generation/image-generation-service.ts
```diff
@@ -24,8 +26,15 @@ export async function getAvailableImageModelsForFederalState({
...
-  return imageModels.find((m) => m.name === DEFAULT_IMAGE_MODEL) ?? imageModels[0];
+export async function getDefaultImageModel({
+  imageModels,
+  federalStateId,
+}: {
+  imageModels: LlmModelSelectModel[];
+  federalStateId: string;
+}): Promise<LlmModelSelectModel | undefined> {
+  return (
+    (await dbGetModelByRoleAndFederalStateId({ role: 'default-image', federalStateId })) ??
```
rlaerm: ImageGenerationService should not use the db functions of module LlmModelService directly. It should always access the data via the service instance because that is where the business logic, validation, error handling, etc. is located.

packages/shared/src/llm-models/llm-model-service.ts
```diff
@@ -1,41 +1,25 @@
...
-
-/**
- * Get the first text model from a list of models, excluding mistral models.
- * @param models The list of LLM models
- * @returns The first text model or undefined if none found
- */
-export function getFirstTextModel(models: LlmModelSelectModel[]): LlmModelSelectModel | undefined {
-  return models.find(
-    (m) => m.priceMetadata.type === 'text' && !m.name.toLowerCase().includes('mistral'),
+export async function getDefaultModel({
```
rlaerm, line 84: The service function calls should have some comments even if the logic is trivial because this functions are meant to be reused all over the code. And exactly those methods should have unit tests.

recursive-hub (approved):

PR #1210, opened 2026-07-29
Feat/td 1449 preset url admin view

rlaerm:


packages/shared/src/utils/url.ts
```diff
@@ -1,3 +1,8 @@
+export const regexes = {
+  domainOnly: /^(?!:\/\/)([a-zA-Z0-9-_]+\.)+[a-zA-Z]{2,11}?$/,
```
AsamMax: Why 11 as the maximum length? ICANN specifies them to be 2-63 chars long. Of course in practice, it's quite esoteric TLDs that get excluded by this like .international or .construction. I'd recommend either 63 (technical limit) or 18 (longest in practice)

apps/admin/src/app/ais-chat-app/url-presets/UrlPresetsListView.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const updatedPreset = await updateUrlPresetAction(presetId, presetData);
+    if (updatedPreset.success) {
+      setUrlPresets((prev) => prev.map((p) => (p.id === presetId ? updatedPreset.value : p)));
+    } else {
+      toast.error(updatedPreset.error.message);
+      await loadUrlPresets();
+    }
+  }
+
+  async function handleDeleteUrlFromPreset(presetId: string, url: string): Promise<void> {
```
AsamMax, line 96: Does this function still work if users rapidly click on the "delete" button? It looks like it could collide.

apps/admin/src/app/ais-chat-app/url-presets/UrlPresetsListView.tsx
```diff
@@ -0,0 +1,186 @@
...
+  };
+
+  useEffect(() => {
+    void loadUrlPresets();
+  }, []);
+
+  async function handleNewUrlPreset(): Promise<void> {
+    const newUrlPreset: UrlPresetInsert = {
+      name: `Neues Webseitenpaket (${urlPresets.length + 1})`,
+      orderNumber: urlPresets.length + 1,
```
AsamMax, line 55: This could cause collisions after deletion, right?

If we already have 3 presets [1,2,3]
Then delete number 2 [1,3]
Then create a new one [1,3,3]

AsamMax (approved):

apps/admin/src/app/ais-chat-app/url-presets/UrlPresetsListView.tsx
rlaerm: Yes, that is right, but not a problem. Both, orderNumber and name are not unique and it is only the admin interface so fine to me.

packages/shared/src/utils/url.ts
rlaerm: Good point. I will change that.

apps/admin/src/app/ais-chat-app/url-presets/UrlPresetsListView.tsx
rlaerm: Yeah, the first one will succeed and the others run into an error. I can disable the button while an update/delete is in progress.

PR #1217, opened 2026-08-03
Add usage estimation to Bifrost provider

AsamMax:


packages/ai-core/src/chat/providers/bifrost.ts
```diff
@@ -128,17 +139,17 @@ export function constructBifrostTextGenerationFn(model: AiModel): TextGeneration
...
 
     const usage = response.usage;
-
-    if (!usage) {
-      throw new AiGenerationError('No usage data returned from Bifrost');
-    }
+    const calculatedUsage = calculateCompletionUsage({
+      messages,
+      modelMessage: { role: 'assistant', content: text },
+    });
```
TobiasWallura-xitaso: Should this be done in an `if`, so we skip calculation, when we already have the usage?

PR #1219, opened 2026-08-03
TD-1465: filter tool messages for export

stschaller-xitaso:


packages/shared/src/conversation/conversation-service.ts
```diff
@@ -152,6 +158,10 @@ export async function getConversationMessageForExport({
     throw new NotFoundError('Conversation message not found');
   }
 
+  if (isToolResultMessage(message)) {
```
recursive-hub: This should not be necessary since the tool call and tool result messages are not displayed in the ui

packages/shared/src/conversation/conversation-service.ts
```diff
@@ -13,6 +13,12 @@ import { ConversationMessageModel, ConversationModel } from '@shared/db/types';
 import { ForbiddenError, NotFoundError } from '@shared/error';
 import { dbGetCharacterById } from '@shared/db/functions/character';
 
+function isToolResultMessage(message: ConversationMessageModel): boolean {
```
recursive-hub: In apps/chat-bot/src/types/chat.ts in the function toUIMessages() we already filter out tool call and tool result messages with this filter:
.filter((m) => m.role !== 'tool' && !m.toolCalls?.length)

Maybe you can use toUIMessages in the export or extract some common parts in another way?

packages/shared/src/conversation/conversation-service.ts
stschaller-xitaso: But would we not still want a server side inforcement (for direct API calls or similar?)

recursive-hub: Ok better safe than sorry :D

recursive-hub (approved):

PR #1221, opened 2026-08-03
TD-1467: Default Bifrost env vars for local dev via .env.development

TobiasWallura-xitaso:
- Adds a committed `.env.development` in `apps/admin` and `apps/chat-bot` with working Bifrost gateway defaults for local development, so `.env.local` no longer needs to set them manually. Next.js loads these files automatically, with `.env.local` overriding any value present in both.
- `apps/api` (Fastify) has no built-in env file loading, so `load-env.ts` now replicates the same precedence manually: it loads `.env.${NODE_ENV}` first, then `.env.local` with `override: true` so it always wins.
- Also switches the deprecated `z.string().url()` to `z.url()` in `apps/admin`/`packages/ai-core` env schemas.
- Configures the e2e workflow's Bifrost container with admin credentials and an API key (previously ran with no auth configured), and moves the common Bifrost env vars to the job-level `env:` section with a header, matching the existing Crawl4AI/Xberg/S3 sections.

Jira: TD-1467

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -72,9 +72,14 @@ jobs:
       # Disable telemetry
       NEXT_TELEMETRY_DISABLED: 1
       TURBO_TELEMETRY_DISABLED: 1
+      # Crawl4AI
       CRAWL4AI_API_TOKEN: ais-chat-crawl4ai-token
+      # Xberg
       XBERG_URL: http://localhost:8000
       XBERG_MAX_UPLOAD_SIZE_MB: '50'
+      # Bifrost
+      BIFROST_BASE_URL: http://localhost:8089/openai/v1
+      BIFROST_API_KEY: sk-bf-example-key
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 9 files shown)


.github/workflows/e2e.yml
```diff
@@ -162,23 +171,11 @@ jobs:
...
-          elapsed=0
-          until wget -qO- http://localhost:8089/api/providers > /dev/null 2>&1; do
-            if [ $elapsed -ge $timeout ]; then
-              echo "Bifrost did not become ready within ${timeout}s"
-              exit 1
-            fi
-            echo "$(date) - waiting for Bifrost to be ready"
-            sleep 1
-            elapsed=$((elapsed + 1))
-          done
```
stschaller-xitaso, line 180: Why do we not need this anymore?

stschaller-xitaso (approved):

.github/workflows/e2e.yml
TobiasWallura-xitaso: Effectively this is not needed, as bifrost is already up and running here. I removed it, to keep the pipeline short.
The bifrost container takes around ~15sec to start. The setup steps between starting the container and the seeding here take around ~60sec.

PR #1224, opened 2026-08-05
Feat/td 1456 web search edit view

rlaerm:


apps/chat-bot/src/components/custom-chat/web-search/web-search-url-preset-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+          disabled={isFullyAdded || isLimitReached}
+        >
+          <PlusIcon />
+          {t('presets-add-preset-button')}
+        </Button>
+      </div>
+
+      <CollapsibleContent>
+        <ul className="flex flex-wrap gap-2 my-4">
+          {preset.urls.toSorted().map((url) => (
```
stschaller-xitaso, line 54: toSorted might not be supported by all browsers, this is the current support:

Image

If that is fine for us, we can leave it, otherwise we might use this to support older browsers?
```suggestion
          {Array.from(preset.urls).sort().map((url) => (
```

apps/chat-bot/src/components/custom-chat/web-search/web-search-selected-websites.tsx
```diff
@@ -0,0 +1,82 @@
...
+            return (
+              <div key={presetName} className="mt-6">
+                <div className="text-xs uppercase text-muted-foreground tracking-wide">
+                  {presetName === UNASSIGNED_PRESET_NAME
+                    ? t('presets-title-unassigned')
+                    : presetName}
+                </div>
+                <ul className="flex flex-wrap gap-2 mt-3">
+                  {websites.map((website, index) => (
+                    <li key={presetName + index}>
```
stschaller-xitaso, line 61: Index keys in React are usually problematic, because they are unstable and lead to different [(issues)](https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/), since this list can change by adding or removing links I would suggest using the (stable) website/url as key or similar (same as in UrlPresetsListView.tsx)

apps/admin/src/app/ais-chat-app/url-presets/UrlPresetsListView.tsx
```diff
@@ -164,21 +164,18 @@ export function UrlPresetsListView() {
...
-                      <Button
-                        onClick={() => handleDeleteUrlFromPreset(preset.id, url)}
-                        variant="ghost"
-                        size="icon-sm"
-                      >
-                        <TrashSimpleIcon data-icon="inline-end" />
-                      </Button>
-                    </Chip>
+                    <Chip
+                      key={`url_${index}`}
```
stschaller-xitaso, line 174: See comment below (I know you did not introduce the index key here, but we should still change it, as its an anti-pattern):
we should use stable url as key instead or anything similar that stays stable

packages/shared/src/web-search/url-presets/url-preset-admin-service.ts
stschaller-xitaso, line 16: Not part of this PR, but currently we do not have any server side validation for correct domain only client side before adding it to the DB. But client validation could be bypassed, should we add server side validation?

stschaller-xitaso (approved):

apps/chat-bot/src/components/custom-chat/web-search/web-search-selected-websites.tsx
rlaerm: You are right, thx.

apps/chat-bot/src/app/(authed)/(chat-bot)/assistants/editor/[assistantId]/assistant-edit.tsx
```diff
@@ -450,11 +457,14 @@ export function AssistantEdit({
           />
 
           {isWebSearchAvailable && (
-            <CustomChatWebSearch
+            <CustomChatWebSearchEditView
               control={control}
               onCheckedChange={() => {
                 void flushAutoSave();
               }}
```
Comment (TobiasWallura-xitaso) [inline apps/chat-bot/src/app/(authed)/(chat-bot)/assistants/editor/[assistantId]/assistant-edit.tsx:464]: The other edit views only have `onChange`.

apps/chat-bot/src/components/entity-overview/entity-overview.tsx
```diff
@@ -223,22 +216,13 @@ export default function EntityOverview({
...
-                      aria-label={`${pill.label}-Filter zurücksetzen`}
-                    >
-                      <XIcon className="size-3" aria-hidden="true" />
-                    </button>
-                  ) : null}
-                </span>
+                  size="sm"
+                  label={pill.label}
+                  onDelete={() => onRemoveFilter(pill.group, pill.value)}
+                  ariaDeleteLabel={`${pill.label}-Filter zurücksetzen`}
```
TobiasWallura-xitaso: unrelated, but we should introduce an i18n key for the aria label.

packages/shared/src/web-search/url-presets/url-preset-service.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Retrieves all URL presets from the database, ordered by orderNumber and name.
+ *
+ * @returns array of UrlPreset objects
+ */
+export async function getAllUrlPresets(): Promise<UrlPreset[]> {
+  const presets = await db
+    .select()
+    .from(urlPresetTable)
+    .orderBy(urlPresetTable.orderNumber, urlPresetTable.name);
```
TobiasWallura-xitaso, line 14: I assume we won't have many changes in the preset table, so we could cache it.
```suggestion
    .orderBy(urlPresetTable.orderNumber, urlPresetTable.name)
    .$withCache();
```

TobiasWallura-xitaso (approved):

apps/chat-bot/src/app/(authed)/(chat-bot)/assistants/editor/[assistantId]/assistant-edit.tsx
Comment (rlaerm) [reply apps/chat-bot/src/app/(authed)/(chat-bot)/assistants/editor/[assistantId]/assistant-edit.tsx:464]: I missed that, thx.

apps/chat-bot/src/components/entity-overview/entity-overview.tsx
rlaerm: done

packages/shared/src/web-search/url-presets/url-preset-service.ts
rlaerm: I am not 100% convinced about the caching. If someone adds a new preset they will check it very soon. This could lead to a wrong bug report. And there is already an optimization on client side in `useUrlPresets`.

PR #1228, opened 2026-08-05
MUNDO Filters

recursive-hub:


apps/chat-bot/src/configuration-text-inputs/const.ts
```diff
@@ -29,6 +29,45 @@ export const MUNDO_SEARCH_TITLE_LENGTH_LIMIT = 200; // Maximum characters kept f
...
+  'Sport',
+  'Arbeitslehre',
+  'Gesundheit',
+  'Interkulturelle Bildung',
+  'Medienpädagogik',
+  'Pädagogik',
+  'Psychologie',
+  'Sucht und Prävention',
+  'Verkehrserziehung',
+] as const;
```
AsamMax, line 70: Do we have a way to load these options dynamically? So we don't have to push code changes when mundo changes something

AsamMax (approved):

apps/chat-bot/src/configuration-text-inputs/const.ts
recursive-hub: Yes there is a way. I would leave it as is until we have a few more cases where a prefetch mechanism makes sense.

PR #1270, opened 2026-08-18
save and restore image size with conversation message

rlaerm:


packages/shared/src/utils/chat.ts
```diff
@@ -35,3 +35,9 @@ export type ImageStyle = {
 
 export const imageStyleTypeSchema = z.enum(['none', 'photorealistic', 'cartoon']);
 export type ImageStyleType = z.infer<typeof imageStyleTypeSchema>;
+
+export const imageAspectRatioPresetSchema = z.enum(['quadratic', 'landscape', 'portrait']);
+export type ImageAspectRatioPreset = z.infer<typeof imageAspectRatioPresetSchema>;
+
+export const imageSizeSchema = z.union([z.literal('auto'), z.string().regex(/^\d+x\d+$/)]);
```
TobiasWallura-xitaso: we probably want to exclude `auto` from the imageSizeSchema.

TobiasWallura-xitaso (approved):

packages/shared/src/utils/chat.ts
rlaerm: I decided to delete the schema for now because it was a left over of storing the size in the conversation_message table. For using it in the admin ui I will create the schema again.

PR #1272, opened 2026-08-18
TD-1478: Adopt Cache Components in chat-bot and admin

TobiasWallura-xitaso:
Enables `cacheComponents: true` in both `chat-bot` and `admin`, dropping the redundant `experimental.useCache` flag.

- Removes `export const dynamic = 'force-dynamic'` across both apps — every route is dynamic by default under Cache Components now.
- Every route that blocks on request-time data (auth/session reads, dynamic params) is opted out of instant navigation via `export const instant = false`, found by running both apps with `pnpm dev` and checking the dev overlay plus the `next build` route table.
- `chat-bot`'s home page (`/`) generated a UUID before any request-time read, which Next tried to bake into the static shell; reordered the auth read to happen first so the UUID generation runs in an already-dynamic context, matching the pattern used by sibling "new conversation" pages.
- Fixes the assistants/characters/learning-scenarios overview lists staying stale after creating or editing an entity: `usePersistedOverviewFilter`'s fetch guard was blocking a refetch when the route was restored from Next's client-side segment cache.

### DIFF
--- a/apps/admin/next.config.ts
+++ b/apps/admin/next.config.ts
@@ -25,9 +25,9 @@ const baseNextConfig: NextConfig = {
   // https://nextjs.org/docs/app/api-reference/config/next-config-js/output#automatically-copying-traced-files
   output: 'standalone',
   reactCompiler: true,
+  cacheComponents: true,
   productionBrowserSourceMaps: !isDevBuild,
   experimental: {
-    useCache: true,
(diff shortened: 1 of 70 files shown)


TobiasWallura-xitaso:
`cacheComponents` still have several open issues, which we will not solve now. -> closing and keeping `experimental.useCache` for now.

PR #1277, opened 2026-08-19
TD-1478: Resolve instant-navigation warnings by Suspense-wrapping session reads

TobiasWallura-xitaso:
Follows up on #1272 to resolve the dev-time instant-navigation warnings that Cache Components flags for every route reading the user session, per Next's [authentication-with-cache-components guide](https://nextjs.org/docs/app/guides/authentication-with-cache-components).

- Cache session reads with `"use cache: private"` (`requireAuth()`, `getUser()`, `getMaybeUser()`, `getMaybeSession()`) so they're cached in the browser instead of blocking.
- Split the shared chat-bot layout and every authed leaf route (assistants, characters, learning-scenarios, image-generation, home, `d/[conversationId]`, `custom/editor`, `top-up`) into a thin wrapper + an async component behind `<Suspense>`, so the session read no longer blocks outside a boundary.
- Same treatment for `/login`, `/logout`, and the public `/ua/*` shared-chat dialog routes, which had no Suspense boundary at all and surfaced the same warning via the root layout.
- Along the way, fixed a pre-existing bug: the assistants/characters/learning-scenarios overview lists stayed stale after creating or editing an entity and navigating back via a soft (client-side) navigation, because a fetch-guard in `usePersistedOverviewFilter` blocked a legitimate refetch when Next restored the route from its client-side segment cache.

Root layout's own session + theming read is intentionally left as a documented `instant = false` Block (not in scope here).

### DIFF
--- a/apps/chat-bot/src/app/(authed)/(chat-bot)/assistants/[assistantId]/page.tsx
+++ b/apps/chat-bot/src/app/(authed)/(chat-bot)/assistants/[assistantId]/page.tsx
@@ -6,6 +6,8 @@ import { AssistantView } from './assistant-view';
 import { DefaultPageLayout } from '@/components/layout/default-page-layout';
 import { type Metadata } from 'next';
 import { getTranslations } from 'next-intl/server';
+import { Suspense } from 'react';
+import { Spinner } from '@ui/components/spinner';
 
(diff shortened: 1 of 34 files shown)


TobiasWallura-xitaso:
Closing, will be revisited later.

PR #1279, opened 2026-08-19
Remove non-agentic stream

recursive-hub:


packages/shared/src/db/schema.ts
```diff
@@ -231,7 +231,6 @@ export const federalStateFeatureTogglesSchema = z.object({
   isCustomGptEnabled: z.boolean().default(true),
   isShareTemplateWithSchoolEnabled: z.boolean().default(true),
   isSharedPageLocaleDetectionEnabled: z.boolean().optional(),
-  isAgenticChatEnabled: z.boolean().optional(),
```
TobiasWallura-xitaso, line 234: Should we add a db migration to remove the property? Otherwise we will keep obsolete data in the database and our typescript types will drift from the actual data (which could become an issue in future, right now we just cast and ignore additional properties).

TobiasWallura-xitaso (approved):

packages/shared/src/db/schema.ts
recursive-hub: I create a ticket, lets discuss this, since this is a recurring issue.

PR #1290, opened 2026-08-21
TD-1494: add size param for image generation to api

stschaller-xitaso:


packages/ai-core/src/images/providers/ionos.ts
```diff
@@ -18,12 +18,13 @@ export function constructIonosImageGenerationFn(llmModel: AiModel): ImageGenerat
...
     prompt,
     model,
+    options,
   }: Parameters<ImageGenerationFn>[0]) {
     const result = await client.images.generate({
       model,
       prompt,
       n: 1,
-      size: '1024x1024',
+      size: options?.size ?? '1024x1024',
```
recursive-hub: Do we need to handle the providers, if we use bifrost? I think we will remove all provider specific code in the current sprint anyway.

TobiasWallura-xitaso: 👍

stschaller-xitaso: Yes, I asked Max about it, because I was also unsure and we landed on adding it for the other providers, because that code should technically still work without bifrost, but since it leads to other issues I reverted that commit, since it is not worth the effort just for having backwards compatibility that is probably not needed anyway

TobiasWallura-xitaso (approved):

PR #1295, opened 2026-08-25
TD-1435: Add admin delete/restore for templates

brueckb:
Adds an admin toggle to soft-delete/restore global templates (character, assistant, learning-scenario) from the templates admin UI, and excludes deleted templates from all browse/listing queries by default via a new `includeDeleted` option on the shared DB lookup functions.

Existing conversations keep working with a deleted template thanks to new `*ForExistingConversation` lookups that join against the conversation table instead of relying on the default (non-deleted) filter, and `chat-service.ts` now validates that the character/assistant/learning-scenario id on an incoming request matches what's stored on the conversation. Web search config resolution was also updated to use the already-loaded entity directly instead of doing a second DB lookup, including in the invite-code based shared chat flows.


apps/admin/src/app/ais-chat-app/templates/[templateType]/[templateId]/actions.ts
Comment (TobiasWallura-xitaso) [inline apps/admin/src/app/ais-chat-app/templates/[templateType]/[templateId]/actions.ts:16]: For consistency we should use `runServerAction` for other actions as well. probably everywhere in the admin app (as a separate tech debt task).

apps/chat-bot/src/app/api/v1/admin/characters/[characterId]/route.ts
Comment (TobiasWallura-xitaso) [inline apps/chat-bot/src/app/api/v1/admin/characters/[characterId]/route.ts:1]: Is this api route still used? I assume we can remove it.

apps/chat-bot/src/app/api/chat/build-tools.ts
```diff
@@ -13,6 +13,7 @@ type BuildToolsParams = {
   learningScenarioId?: string;
   assistantId?: string;
   conversationId?: string;
+  webSearchEntity?: WebSearchModel;
```
TobiasWallura-xitaso: Does the new `webSearchEntity` replace the existing `characterId`, `learningScenarioId` and `assistantId`?

apps/chat-bot/src/app/api/chat/websearch.ts
```diff
@@ -49,16 +49,27 @@ export async function resolveWebSearchConfig({
   characterId,
   learningScenarioId,
   assistantId,
+  entity,
 }: {
   user: UserAndContext;
   characterId?: string;
   learningScenarioId?: string;
   assistantId?: string;
+  entity?: WebSearchModel;
```
TobiasWallura-xitaso: Same as other comment: does `entity` replace the other id props?

packages/shared/src/db/functions/assistants.ts
TobiasWallura-xitaso, line 1: That's a lot of changes for handling deleted entities. Would it also be ok to exclude soft-deleted entities in all queries by default? Conversations with soft-deleted entities would then be unusable, but that might be ok.
What is the expected behavior for conversations with soft-deleted entities? e.g. should they be readonly or can they still be used for chat.

TobiasWallura-xitaso, line 291: Deleting entities by the user still deletes it from db. We could change that to soft-delete. What is the expected behavior?

TobiasWallura-xitaso (approved):

apps/admin/src/app/ais-chat-app/templates/[templateType]/[templateId]/actions.ts
Comment (brueckb) [reply apps/admin/src/app/ais-chat-app/templates/[templateType]/[templateId]/actions.ts:16]: added to the tech debt list.

apps/chat-bot/src/app/api/chat/build-tools.ts
brueckb: no it does not, sorry for the misleading name. `webSearchEntity` is just used as a short-circuit, otherwise the `*id` is still needed for loading from the db. I renamed `webSearchEntity` to `webSearchSettings` for more clarity.

apps/chat-bot/src/app/api/v1/admin/characters/[characterId]/route.ts
Comment (brueckb) [reply apps/chat-bot/src/app/api/v1/admin/characters/[characterId]/route.ts:1]: currently used for load tests. i will leave it in until we refactor those.

PR #1297, opened 2026-08-25
TD-1519: Add qalc math calculation tool

AsamMax:
## Summary

- Add a sandboxed standalone qalc/libqalculate service built from pinned official release 5.12.0.
- Add the feature-gated `math_calculate` chatbot tool with qalc syntax guidance and bounded response handling.
- Add admin toggle, Compose wiring, failure handling, and focused tests.

## Verification

- `pnpm format:check`
- `pnpm lint`
- `pnpm check-types`
- `pnpm test`
- Docker image build and real qalc HTTP smoke tests
- Both Compose configurations validated
- Local chat and admin started successfully
- Enabled QALC for `DE-BY` through the admin UI
- GPT-5 mini called `math_calculate` with `factor(x^12 - 1)`
- Persisted tool result and final assistant response matched
- Desktop and mobile browser regression completed with no console errors during the successful flow


apps/chat-bot/src/app/api/chat/calculator.ts
```diff
@@ -0,0 +1,134 @@
+import { z } from 'zod';
+import { env } from '@/env';
+
+export const CALCULATOR_MAX_EXPRESSION_LENGTH = 4_096;
+export const CALCULATOR_MAX_OUTPUT_LENGTH = 16_000;
+export const CALCULATOR_TIMEOUT_MS = 5_000;
+
+const CALCULATOR_STATUSES = [
```
recursive-hub: this was defined also in services/calculator/src/types.ts

services/calculator/src/server.ts
```diff
@@ -0,0 +1,130 @@
...
+  overload: 429,
+  timeout: 504,
+  crashed_worker: 502,
+  malformed_output: 502,
+  upstream_failure: 503,
+  internal_failure: 500,
+};
+
+// Expose GET /healthz for probes and POST /v1/calculate for bounded evaluations.
+function send(response: ServerResponse, result: Result): void {
```
recursive-hub: As far as I understand it, this service is basically one POST endpoint that spawns a qalc subprocess and returns the result. Why do we need a custom HTTP server and a worker pool instead of a simple framework like Fastify and maybe p-limit?

services/calculator/src/worker.ts
```diff
@@ -0,0 +1,172 @@
+import { spawn } from 'node:child_process';
```
recursive-hub: This file contains quite a bit of low-level subprocess code. Could we simplify this by using a library like execa here?

recursive-hub (approved):

services/calculator/src/server.ts
```diff
@@ -0,0 +1,110 @@
+import fastify, { type FastifyError, type FastifyInstance } from 'fastify';
```
recursive-hub, line 1: can we use fastify even more here?

PR #1301, opened 2026-08-26
ci(e2e): attach app and container logs to each Playwright test

TobiasWallura-xitaso:
E2E failures currently require manually correlating a wall of app/container logs against test timing. This surfaces the relevant logs per test instead.

- Timestamp the e2e app log (UTC) and stream container logs with timestamps to files during the run.
- Add a Playwright reporter that slices both log files to each test's time window and attaches them, so per-test server + container logs show up in the HTML report.
- Emit all app log levels to stdout in CI (previously only errors surfaced).

Slices are time-windowed and approximate: with 3 workers concurrent tests may share some lines, and second-resolution app timestamps use a tolerance padding.


.github/workflows/e2e.yml
```diff
@@ -144,9 +144,12 @@ jobs:
           cp -a .next/static .next/standalone/apps/chat-bot/.next/
           cp -a public .next/standalone/apps/chat-bot/
           cd .next/standalone
-          node ./apps/chat-bot/server.js > /tmp/ais-chat-app.log 2>&1 &
+          node ./apps/chat-bot/server.js 2>&1 \
+            | gawk '{ print strftime("%Y-%m-%dT%H:%M:%SZ", systime(), 1), $0; fflush() }' \
+            > /tmp/ais-chat-app.log &
```
TobiasWallura-xitaso: Add a short comment why the gawk step is done (prepending log lines with timestamps, so they can be correlated with playwright tests).

.github/workflows/e2e.yml
```diff
@@ -164,6 +167,11 @@ jobs:
             elapsed=$((elapsed + 1))
           done
 
+      # Stream container logs (with timestamps) to a file so the Playwright reporter can slice
+      # them per test; the same file is printed at the end of the job.
```
TobiasWallura-xitaso:
```suggestion
      # Stream container logs (with timestamps) to a file so the Playwright reporter can correlate them with tests.
```

packages/shared/src/logging/logging.ts
```diff
@@ -8,6 +8,11 @@ const logLevels = logLevelOrder.slice(
   1 + logLevelOrder.indexOf(env.NEXT_PUBLIC_SENTRY_LOG_LEVEL),
 );
 
+// Write logs to stdout in development, or when explicitly enabled (e.g. e2e CI) so
+// prod-parity builds can still surface logs without changing NODE_ENV. LOG_TO_STDOUT is a
+// server-only env var, so guard the access to avoid throwing in client bundles.
```
TobiasWallura-xitaso, line 13:
```suggestion
// Write logs to stdout in development, or when explicitly enabled (e.g. e2e CI) so
// prod builds can still surface logs without changing NODE_ENV. LOG_TO_STDOUT is a
// server-only env var, so guard the access to avoid throwing in client bundles.
```

AsamMax (approved):

PR #1302, opened 2026-08-26
TD-1486: add editor role for admin

stschaller-xitaso:


apps/admin/src/app/api/auth/logout-callback/route.ts
```diff
@@ -7,15 +7,15 @@ const SECURE_SESSION_COOKIE_NAME = `__Secure-${SESSION_COOKIE_NAME}`; // Used wh
 
 /**
  * This route is called by the IDP after logout.
- * We clear the session cookies and redirect to the login page.
+ * We clear the session cookies and redirect to the application root.
  * If the session cookie is bigger than 4 kb, the cookie might be split into multiple cookies.
```
stschaller-xitaso, line 11: Admin does not have a login page, instead it is redirected to /api/auth/signin for Keycloak sign in. Redirecting to /login causes 404 error

recursive-hub (approved):

PR #1304, opened 2026-08-27
Handle model change + Image generation tests

recursive-hub:


apps/chat-bot/src/components/image-generation/image-generation-result.tsx
```diff
@@ -26,7 +26,13 @@ export function ImageGenerationResult({
           <h3 className="text-xs text-gray-700">{tImageGeneration('attached-files-label')}</h3>
           <div className="flex flex-row gap-2 overflow-auto mt-2 mb-5">
             {attachedFiles.map((file) => (
-              <MessageImageAttachment key={file.id} file={file} width={56} height={56} />
+              <MessageImageAttachment
+                key={file.id}
+                file={file}
+                width={480}
+                height={160}
+                className="h-20 w-auto max-w-60 max-h-none object-cover aspect-auto"
```
stschaller-xitaso, line 34: Why do we need the classnames here? Just asking because I am currently working on the issues with SVGs and multiple image upload for scaled images and have to adapt the classnames of the MessageImageAttachment. Want to be sure what to look out for/not to break something else

stschaller-xitaso (approved):

apps/chat-bot/src/components/image-generation/image-generation-result.tsx
recursive-hub: This is for keeping the height fixed and computing the width automatically, keeping the aspect ratio.

PR #1330, opened 2026-09-09
TD-1530: Add configurable chat safety checks

AsamMax:
Adds configurable input safety checks to first-party and API chat flows. Safety is selected through the Admin-managed global static model configuration, runs once before generation, excludes system/tool messages, and forwards text plus image inputs to the direct Google Vertex Llama Guard provider. Invalid images are filtered and reported without breaking chat; assistant-output moderation is intentionally deferred.


apps/api/src/routes/(app)/v1/chat/completions/post.ts
```diff
@@ -59,6 +60,13 @@ export async function handler(request: FastifyRequest, reply: FastifyReply): Pro
   }
 
   const body = requestParseResult.data;
+  const safetyModel = (await dbGetModelsByApiKeyId({ apiKeyId: apiKey.id })).find(
+    (model) => model.priceMetadata.type === 'safety',
+  );
+  if (!safetyModel) {
+    reply.status(500).send({ error: 'No safety model configured for this API key' });
+    return;
+  }
```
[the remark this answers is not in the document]
AsamMax: Addressed: API chat intentionally resolves safety from API-key mappings because it uses the separate API database.

packages/ai-core/src/safety/providers/google.ts
```diff
@@ -1,167 +1,150 @@
...
+import type { AiModel, SafetyCheckFn, SafetyImage, SafetyResult } from '../types';
+
+const MAX_IMAGES = 4;
+const MAX_IMAGE_URL_LENGTH = 10 * 1024 * 1024;
+
+const responseSchema = z.object({
+  predictions: z.object({
+    choices: z.array(
+      z.object({
+        message: z.object({ content: z.string() }),
```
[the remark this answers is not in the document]
AsamMax: Addressed: the current parser supports both Vertex response shapes.

packages/ai-core/src/safety/providers/google.ts
```diff
@@ -1,167 +1,160 @@
...
-      return { result: validateSafetyResult(result) };
+      return parseSafetyResult(await response.json(), model.id);
     } catch (error) {
-      throw new AiGenerationError(
-        `Google Vertex AI Safety request failed: ${error instanceof Error ? error.message : String(error)}`,
-      );
+      Sentry.captureException(error, {
+        tags: { provider: 'google', operation: 'safety_check' },
+      });
+      return { safe: true };
```
stschaller-xitaso, line 157: Why do we return safe: true here for errors? Wouldn't that lead to any errors like malformed responses, unauthorized etc to be swallowed and classified as safe?

brueckb (approved):

packages/ai-core/src/safety/providers/google.ts
AsamMax: Yes it would, but the alternative is that no requests work if the google authentication breaks. I've also asked the PO, she also sees it this way

packages/ai-core/src/safety/index.ts
```diff
@@ -1,22 +1,45 @@
...
-  if (!(await hasAccessToModel(apiKeyId, model))) {
-    throw new InvalidModelError(`API key does not have access to the safety model: ${model.name}`);
+export function interpretSafetyResult(result: SafetyResult): void {
+  if (result.safe) {
+    return;
   }
 
-  // TODO: Add image inputs once the deployed Llama Guard endpoint supports them.
-  return checkSafety(model, text);
+  throw new ResponsibleAIError('Input was blocked by the safety model');
```
TobiasWallura-xitaso, line 14: How is this error shown to the user?

packages/ai-core/src/safety/providers/google.ts
```diff
@@ -1,167 +1,160 @@
...
 
-      return { result: validateSafetyResult(result) };
+      return parseSafetyResult(await response.json(), model.id);
     } catch (error) {
-      throw new AiGenerationError(
-        `Google Vertex AI Safety request failed: ${error instanceof Error ? error.message : String(error)}`,
-      );
+      Sentry.captureException(error, {
+        tags: { provider: 'google', operation: 'safety_check' },
+      });
```
TobiasWallura-xitaso, line 158: use `logError` instead of calling Sentry directly

packages/ai-core/src/safety/providers/google.ts
```diff
@@ -1,167 +1,160 @@
...
-        ],
-      };
-
-      const requestUrl = `https://${endpointHost}/v1/projects/${settings.projectId}/locations/${settings.location}/endpoints/${endpointId}:predict`;
-
-      const response = await fetch(requestUrl, {
+      const images = filterImages(messages.flatMap((message) => message.images ?? []));
+      const imageParts = images.map((image) => ({
+        type: 'image_url' as const,
+        image_url: { url: image.url },
```
TobiasWallura-xitaso, line 124: Does this handle both S3 presigned URLs and inline base64 encoded images?

packages/ai-core/src/safety/providers/google.ts
```diff
@@ -1,167 +1,160 @@
...
-    return normalized;
+function filterImages(images: SafetyImage[]): SafetyImage[] {
+  const validImages = images.filter(isValidImage);
+  const droppedImageCount = images.length - Math.min(validImages.length, MAX_IMAGES);
+
+  if (droppedImageCount > 0) {
+    Sentry.captureMessage('Safety image attachments filtered', {
+      level: 'warning',
+      extra: { droppedImageCount, maximumImageCount: MAX_IMAGES },
+    });
```
TobiasWallura-xitaso, line 80: Use our logging utilities

TobiasWallura-xitaso: -> probably not used because of circular project dependencies. create tech debt or implement as a separate PR

packages/ai-core/src/safety/providers/google.ts
```diff
@@ -1,167 +1,160 @@
...
-  if (typeof prediction === 'string') {
-    return prediction;
-  }
-  if (!isRecord(prediction) || !Array.isArray(prediction.choices)) {
-    return undefined;
+  if (classification === 'UNSAFE') {
+    return {
+      safe: false,
+      categories: categories
+        .flatMap((category) => category.split(','))
```
TobiasWallura-xitaso, line 48: Why do we split by `,`? Our prompt example defines that categories should be separated by whitespaces.

packages/ai-core/src/chat/agent-loop.ts
```diff
@@ -94,6 +98,17 @@ export function runAgentLoop({
...
     try {
+      if (safetyModelName) {
+        const safetyMessages = messages
+          .filter((message) => message.role === 'user' || message.role === 'assistant')
+          .map((message) => ({
+            role: (message.role === 'user' ? 'user' : 'assistant') as 'user' | 'assistant',
+            content: message.content,
+            images: message.attachments?.filter(isChatImageAttachment),
+          }));
+        await checkInputSafety(safetyModelName, safetyMessages, apiKeyId);
```
TobiasWallura-xitaso, line 109: Do I understand it correctly, that only input is run through the guard model and the generated llm response is not checked?
If so, is that intended behavior?

TobiasWallura-xitaso, line 109: How does cost tracking work for safety checks?

packages/ai-core/src/safety/index.ts
AsamMax: The same way as ResponsibleAIErrors thrown directly by the llm. The red block, and an error message indicating that the message is unsafe.

packages/ai-core/src/chat/agent-loop.ts
AsamMax: Yes, agreed on with PO as a first version. Output can be added later.

AsamMax: The safety model currently doesn't have a price (since it is an hourly billed server and not billed per request or token)

packages/ai-core/src/safety/providers/google.ts
AsamMax: Yes, but good point. I've added tests for this

packages/ai-core/src/safety/providers/google.ts
AsamMax: I added it to tech debt, as it would explode the scope of this feature/ticket.

packages/ai-core/src/safety/providers/google.ts
AsamMax: This is additionally to already being split by whitespace. Sometimes the model had used commas despite the prompt.

PR #1333, opened 2026-09-09
TD-1548: Bill generations that end in an abort

brueckb:
Follow-up to the TD-1548 memory work (#1323, #1324, #1325). Billing was gated entirely on the provider's `finish` event, which every provider yields only after its stream loop completes. An abort makes that loop throw, so the event never arrived and nothing was recorded: no api-key quota via `billTextGenerationUsageToApiKey`, no `conversation_usage_tracking` rows, and a `new-message` analytics event carrying zeros. Before #1324 the stop button did not abort server-side work at all, so this was latent; now that consumer cancel, the queue backlog check, the idle timeout and the 10-minute ceiling all fire aborts automatically, closing a tab mid-generation means unmetered inference.

Providers now catch the abort and emit `finish` instead of throwing, so the single existing billing path keeps working untouched. Google reports cumulative `usageMetadata` on every chunk and Anthropic reports usage through `message_start` and `message_delta`, so both bill real provider numbers on abort — the Anthropic events were previously only read via `finalMessage()` at `message_stop`, which an abort never reaches. The OpenAI family and IONOS only report usage in their final chunk, so those fall back to a local tiktoken estimate, flagged with `estimated: true` on `TokenUsage` and counted by a new `estimated_token_usage` metric per model and provider. Note the estimator is biased low: it counts message text only, ignoring image attachments, tool definitions and provider-side overhead.

The agent loop also had two holes that dropped usage independently of the abort work. It returned early on empty text without calling `onComplete`, and `onError` received no usage at all, so a failure in a later agent iteration discarded what earlier iterations had already been billed for. `onError` now carries the accumulated usage, and the three chat services share a `persistUsage` helper called from both the success and failure paths.

One deliberate omission: `persistAssistantMessage` still calls `getChatTitle` on the abort path, so closing the tab on a first message costs one extra auxiliary call. Skipping it would leave the conversation untitled in the sidebar — happy to change it if reviewers prefer.

### DIFF
--- a/apps/chat-bot/src/app/api/character/character-chat-service.ts
+++ b/apps/chat-bot/src/app/api/character/character-chat-service.ts
@@ -1,4 +1,9 @@
-import { TokenPointsExceededError, SharedChatExpiredError, runAgentLoop } from '@ais-chat/ai-core';
+import {
+  TokenPointsExceededError,
+  SharedChatExpiredError,
+  runAgentLoop,
+  type TokenUsage,
+} from '@ais-chat/ai-core';
 import { NotFoundError } from '@shared/error';
 import { createTextStream, encodeChatStreamEvent } from '@/utils/streaming';
 import { getUserAndContextByUserId } from '@/auth/utils';
@@ -195,6 +200,45 @@ export async function sendCharacterMessage({
     imageAttachmentType,
   );
 
+  const persistUsage = async ({
+    usage,
+    priceInCents,
+    modelUsages,
+  }: {
+    usage: TokenUsage;
+    priceInCents: number;
+    modelUsages: Array<{ modelId: string; usage: TokenUsage; priceInCents: number }>;
+  }) => {
+    if (modelUsages.length === 0) {
+      return;
+    }
+
+    // Agentic requests can invoke several models across iterations. Persist each usage
+    // entry separately so pricing and reporting stay associated with the serving model.
+    for (const modelUsage of modelUsages) {
+      await dbUpdateTokenUsageByCharacterChatId({
+        modelId: modelUsage.modelId ?? generationModelId,
+        completionTokens: modelUsage.usage.completionTokens,
+        promptTokens: modelUsage.usage.promptTokens,
+        characterId: character.id,
+        userId: teacherUserAndContext.id,
+        costsInCent: modelUsage.priceInCents,
+      });
+    }
+
+    await sendRabbitmqEvent(
+      constructNewMessageEvent({
+        user: teacherUserAndContext,
+        promptTokens: usage.promptTokens,
+        completionTokens: usage.completionTokens,
+        costsInCent: priceInCents,
+        provider: definedModel.provider,
+        anonymous: true,
+        character,
+      }),
+    );
+  };
+
   runAgentLoop({
     modelSelection,
     apiKeyId,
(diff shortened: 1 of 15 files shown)


apps/chat-bot/src/app/api/chat/chat-service.ts
```diff
@@ -611,8 +625,9 @@ export async function sendChatMessage({
         streamError(error instanceof Error ? error : new Error('Unknown error'));
       }
     },
-    onError: async (error: Error) => {
+    onError: async (error: Error, billedUsage) => {
       await persistEmptyAssistantMessage();
+      await persistUsage(billedUsage);
 
       streamError(error);
```
stschaller-xitaso: streamError(error) is never reached if the methods (persistence) fail, can we add a try/catch/finally block here so error is handled and streamError is actually reached?
```suggestion
      try {
        await persistEmptyAssistantMessage();
        await persistUsage(billedUsage);
      } catch (persistenceError) {
        logError('Error persisting failed agent loop usage:', persistenceError);
      } finally {
        streamError(error);
      }
```

apps/chat-bot/src/app/api/character/character-chat-service.ts
```diff
@@ -206,37 +249,14 @@ export async function sendCharacterMessage({
...
+      await persistUsage({ usage, priceInCents, modelUsages });
 
       done();
     },
-    onError: (error) => {
+    onError: async (error, billedUsage) => {
       logError('Error during character chat streaming:', error);
+      await persistUsage(billedUsage);
+
       streamError(error);
```
stschaller-xitaso: Same here

apps/chat-bot/src/app/api/learning-scenario/learning-scenario-chat-service.ts
```diff
@@ -211,37 +254,14 @@ export async function sendLearningScenarioMessage({
...
+      await persistUsage({ usage, priceInCents, modelUsages });
 
       done();
     },
-    onError: (error) => {
+    onError: async (error, billedUsage) => {
       logError('Error during shared chat streaming:', error);
+      await persistUsage(billedUsage);
+
       streamError(error);
```
stschaller-xitaso: Same here

stschaller-xitaso (approved):

packages/ai-core/src/chat/agentic-stream.ts
```diff
@@ -68,11 +75,19 @@ export async function* generateAgenticStreamWithBilling(
...
           billingModel,
           event.usage,
         );
+
+        if (event.usage.estimated) {
+          estimatedUsageCounter.add(1, {
+            'gen_ai.request.model': billingModel.name,
+            'gen_ai.provider.name': billingModel.provider,
+          });
+        }
```
TobiasWallura-xitaso, line 90: If possible, a (custom) attribute on the Sentry span would be nice as well.

TobiasWallura-xitaso (approved):

packages/ai-core/src/chat/agentic-stream.ts
brueckb: added in `agent-loop.ts` (https://github.com/FWU-DE/ais-chat/pull/1333/changes#diff-6d6a4b59e75327594ff040c018db37095f5a2955b35666eca0469a940428efb3R139)

PR #1335, opened 2026-09-09
TD-1490: display user relevant error messages

stschaller-xitaso:


apps/chat-bot/src/hooks/use-ais-chat.ts
```diff
@@ -184,6 +184,10 @@ export function useAisChat({
               continue;
             }
 
+            if (streamEvent?.type === 'error') {
+              throw deserializeError(streamEvent.error);
```
stschaller-xitaso: Check what if error is from toolcall and model could continue despite error

apps/chat-bot/src/error/get-error-message-by-type.ts
```diff
@@ -2,6 +2,9 @@ import { isKnownAiGenerationError, type KnownAiGenerationError } from '@ais-chat
 
 type ErrorMessageTranslationKey =
   | 'rate-limit-error'
+  | 'provider-rate-limit-error'
+  | 'responsible-ai-error'
+  | 'invalid-model-error'
```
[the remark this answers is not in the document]
stschaller-xitaso: That's okay since these jsons are only used for shared chat locale

apps/api/src/ai-core-adapter/errors.test.ts
```diff
@@ -66,14 +80,14 @@ describe('handleAiCoreError', () => {
...
+  it('does not infer an error type from an AiGenerationError message', () => {
     const reply = createMockReply();
     const handled = handleAiCoreError(
       reply as never,
       new Error('AiGeneration: exceeded its monthly quota'),
     );
     expect(handled).toBe(true);
-    expect(reply.statusCode).toBe(429);
+    expect(reply.statusCode).toBe(500);
   });
```
TobiasWallura-xitaso: Is this test even needed or can we remove it now?

packages/ai-core/src/errors.ts
```diff
@@ -67,7 +67,46 @@ export class RateLimitExceededError extends AiGenerationError {
...
-      return 'name' in error && error.name === 'RateLimitExceededError';
+      return (
+        'name' in error &&
+        (error.name === 'RateLimitExceededError' ||
+          error.name === 'ApiKeyQuotaExceededError' ||
+          error.name === 'ProviderRateLimitExceededError')
+      );
+    }
+    return false;
+  }
```
TobiasWallura-xitaso, line 78: `RateLimitExceededError` is never instantiated directly in production anymore — only the two subclasses are thrown, and `RateLimitExceededError.is()` itself is never called directly (only reached redundantly via `aiGenerationErrorTypes`, where the subclass checks already cover everything). Can this be made `abstract`? That'd enforce picking a specific subtype at every throw site instead of relying on convention. Note the two `new RateLimitExceededError(...)` calls in `errors.test.ts` (lines 52, 59) would need to switch to a concrete subclass.

🤖 Written by AI under human supervision.

apps/chat-bot/messages/de.json
```diff
@@ -40,6 +40,10 @@
     "regenerate-message": "Antwort erneut generieren",
     "image-load-failed": "Bild konnte nicht geladen werden",
     "rate-limit-error": "Das bereitgestellte Tokenvolumen wurde aufgebraucht.",
+    "api-key-quota-error": "Das bereitgestellte Kontingent des Bundeslandes wurde aufgebraucht.",
+    "provider-rate-limit-error": "Das Modell ist derzeit ausgelastet. Bitte versuche es später erneut.",
+    "responsible-ai-error": "Die Anfrage wurde aus Sicherheitsgründen abgelehnt.",
+    "invalid-model-error": "Das ausgewählte Modell ist nicht verfügbar.",
```
TobiasWallura-xitaso, line 46: Only `de.json` gets these keys; `en.json`/`fr.json`/`it.json`/`ar.json` are missing them (pre-existing gap, not caused here). Since these are consumed by shared-chat's `generic-shared-chat.tsx`, can we open a separate ticket to localize error messages there across all locales?

🤖 Written by AI under human supervision.

packages/ai-core/src/errors.ts
```diff
@@ -141,6 +180,88 @@ export class SharedChatExpiredError extends AiGenerationError {
...
+  ) {
+    return new ProviderRateLimitExceededError(message);
+  }
+
+  if (
+    (normalizedCode !== undefined && RESPONSIBLE_AI_ERROR_CODES.has(normalizedCode)) ||
+    normalizedMessage.includes('request was rejected by the safety system') ||
+    normalizedMessage.includes('content policy violation')
+  ) {
+    return new ResponsibleAIError(stripProviderRequestId(message));
```
TobiasWallura-xitaso: Checked this against every current path: direct OpenAI SDK, Anthropic SDK, and Bifrost (which also goes through the `openai` npm client, since Bifrost is OpenAI-API-compatible — see `bifrost.ts`). All of them keep `requestID` as a separate property (from the `x-request-id`/`request-id` header), never appended to `.message`. So this regex has no real trigger path today. Did you see this "Request ID: ..." suffix in an actual provider/Bifrost error, or was it added speculatively alongside the test? If there's no concrete case, suggest dropping `stripProviderRequestId` and its test until we hit a real example.

🤖 Written by AI under human supervision.

TobiasWallura-xitaso (approved):

apps/chat-bot/messages/de.json
stschaller-xitaso: I asked Elena and added it to the current ticket, tbd if it should be done/a separate ticket

packages/ai-core/src/errors.ts
stschaller-xitaso: It was a warning from Copilot that this could leak provider ids, but if they seem to be never in the message it should be fine -> I removed it

PR #1368, opened 2026-09-15
TD-1561: Fix broken shared-chat image attachments after reload

brueckb:
Unauthenticated shared-chat image attachments broke after reload because persisted attachments lost their blob URLs and fell back to the authenticated file route.
This uses a shared-chat-scoped scaled-image route that validates the invite/entity/session/file ownership before reusing the existing Sharp thumbnail service. Restored images point directly to this route; browser caching makes subsequent loads fast, while the first request performs authorization, S3 retrieval, and resizing.
Also hardens shared-file metadata checks and adds focused route/service tests.


apps/chat-bot/src/components/chat/chat-box.tsx
```diff
@@ -103,7 +103,12 @@ export function ChatBox({
         {imageFiles.length > 0 && (
           <div className="flex flex-row gap-2 overflow-auto">
             {imageFiles.map((file) => (
-              <MessageImageAttachment file={file} key={file.id} />
+              // The authenticated scaled-image fallback only works for files owned by the current user (fileMapping).
```
TobiasWallura-xitaso, line 106: Did you consider introducing an unauthenticated scaled-image route instead of using pre-signed URLs?

brueckb: should have been obvious, but i didn't - thanks for the hint! I will look into it

stschaller-xitaso (approved):

apps/chat-bot/src/components/chat/message-image-attachment.tsx
```diff
@@ -14,15 +15,33 @@ type MessageImageAttachmentProps = {
   width?: number;
   height?: number;
   className?: string;
+  /**
+   * Whether falling back to `/api/files/[fileId]/scaled-image` is valid when no
+   * `localUrl` is set. That route only works for files owned by the current
+   * authenticated user, so shared-chat (anonymous) callers must set this to
+   * `false` and wait for a resolved signed URL instead.
+   */
+  allowFallbackUrl?: boolean;
```
TobiasWallura-xitaso: Is this still valid?

brueckb: updated

